PR #12, opened 2026-05-30
fix(chrome): unblock Cerebras blanks in normal inputs

likethemonth:
## Summary

This PR fixes the Chrome setup path where Cerebras appeared configured correctly in the popup self-check, but `_` blanks did not answer in normal page inputs such as the LinkedIn search box.

There were two related problems in the setup/debug path:

1. **Chrome native messaging host did not inherit shell API keys**
   - `opencues set-key cerebras ...` writes `CEREBRAS_API_KEY` to `~/.cues/.env`.
   - The native host is launched by Chrome, not by the user's interactive shell, so it often does not inherit `.zshrc`/shell exports.
   - Result: the popup could show user-pasted keys, but host-pushed runtime config could still miss keys depending on how the browser/host process was launched.

2. **Normal search inputs with `autocomplete="off"` were treated as sensitive**
   - Chrome normal-input mode intentionally refuses password, OTP, payment, and credential-like fields.
   - The prior implementation treated any `autocomplete="off"` field as sensitive.
   - Many normal search boxes, including LinkedIn-style search inputs, use `autocomplete="off"` for UX reasons, so OpenCues could silently refuse to attach even while the self-check only reported `currentTarget: <input>`.
   - Result: the provider/model/key checks passed, but the runtime never attached to the field that should have triggered the blank.

### Chrome host key loading

- Added `.env` loading to `integrations/chrome/host/host.cjs`.
- The host now reads `~/.cues/.env` at startup and again before a scheduled config push.
- Real process env still wins, so explicit environment overrides continue to work.
- `sendHostConfig()` now reruns on file-change scheduling so key changes are pushed live to the extension.

### CLI provider support

- Extended `opencues set-key` provider mapping beyond Groq/OpenAI/Finnhub.
- Added support for:
  - `cerebras` -> `CEREBRAS_API_KEY`
  - `gemini` -> `GEMINI_API_KEY`
  - `anthropic` -> `ANTHROPIC_API_KEY`
  - `openrouter` -> `OPENROUTER_API_KEY`
  - `opencode-zen` -> `OPENCODE_ZEN_API_KEY`
- Added the Cerebras example to the command help.

### Normal input attach behavior

- Changed `autocomplete="off"` handling from a blanket sensitive-field block to contextual handling.
- Introduced `SENSITIVE_AUTOCOMPLETE_OFF_CONTEXT_PATTERN` for higher-signal payment/account/security context around `autocomplete="off"`.
- Ordinary search inputs with `autocomplete="off"` now remain attachable.
- Payment/account-like `autocomplete="off"` fields still get rejected.
- Kept the existing hard blocks for password, OTP, credit-card autocomplete tokens, and name/id credential heuristics.

### Diagnostics

- Moved Chrome content-script `currentTarget` into module scope so the popup diagnostic can report actual OpenCues attachment state.
- Added `attachStatus` to the self-check output.
- Self-check now distinguishes:
  - an active DOM input is focused
  - OpenCues actually attached to it
  - OpenCues refused it because it is unsupported or sensitive
- Exposed the trust gate on `window.__opencuesTrustGate` so the existing diagnostic flag reports accurately.

### Tests and docs

- Added coverage that ordinary `type="search" autocomplete="off"` fields remain attachable.
- Added coverage that `autocomplete="off"` fields in payment context are still rejected.
- Updated Chrome normal-input and security docs to document the contextual `autocomplete="off"` policy.

## Validation

Ran:

```bash
pnpm --filter @opencues/chrome typecheck
pnpm --filter @opencues/chrome test -- --runInBand
pnpm --filter @opencues/chrome test -- src/ambient-context.test.ts src/sensitive-field-docs.drift.test.ts
opencues install chrome
opencues doctor
```

Results:

- Chrome typecheck passed.
- Chrome test suite passed: 155 tests.
- Extension rebuilt successfully at `integrations/chrome/dist`.
- `opencues doctor` reports Chrome extension build output present and Chrome native messaging host installed.
- Local production-path Cerebras fluid blank smoke test for `the inventor of python is _` returned `Guido van Rossum`.

## Manual upgrade path for this fix

After installing this build locally, Chrome users need to:

1. Reload the unpacked OpenCues extension in `chrome://extensions`.
2. Hard-refresh the target tab (`Cmd+Shift+R` / `Ctrl+Shift+R`).
3. Focus the target input again.
4. Re-run self-check and confirm `attachStatus: attached to <input>`.
5. Delete and retype the final `_` so the trusted underscore gate sees a fresh user insertion.

## Security notes

This narrows only the `autocomplete="off"` false-positive. It does not relax the core sensitive-field protections:

- password/OTP/credit-card autocomplete tokens still block attach
- credential-like `name`/`id` values still block attach
- `autocomplete="off"` still blocks when surrounding field/form metadata looks payment/account/security related
- ambient context continues to return `null` for sensitive fields


wkasekende:
Reviewed and added a fixup commit (a46d60f) with three changes:

1. **`host.cjs` — live `.env` rotation now actually works.** The original `if (!process.env[key])` guard meant `loadEnvFile` was a no-op on every reload after the first — `opencues set-key` wouldn't propagate. Added an `ENV_FILE_KEYS` Set so file-sourced keys can be overwritten on reload while shell exports still win first-load precedence. Verified the three precedence cases (file-only load, file rotation, shell-wins) with a standalone node trace.

2. **`opencues-bootstrap.ts` — closed a gap in the `autocomplete="off"` relaxation.** `SENSITIVE_FIELD_NAME_PATTERN` doesn't catch `account-number` / `iban` / `card-number` (it's tuned for password/cvv/otp/secret/token), so a bare `<input name="iban" autocomplete="off">` with no surrounding form/placeholder context would have slipped through. Added the input's own `name` + `id` to the off-context haystack. New test pins all three names.

3. **`opencues-bootstrap.ts` + `content.ts` — narrowed the trust-gate diagnostic to a boolean sentinel.** The original `window.__opencuesTrustGate = trustGate` exposed `noteUnderscoreInsertion` / `reset` — invisible to page scripts today (Chrome isolated world) but would let a MAIN-world execution forge credits and bypass the gate. Sentinel-only (`__opencuesTrustGateInstalled = true`) is strictly narrower; the popup's `trustGateInstalled` field is unchanged.

All 156 tests pass (was 155 — added one); typecheck + build clean. Will merge once CI re-runs.

PR #53, opened 2026-06-02
fix(macos): complete #51 — rewrite Terminal.app double-ESC arrows at stdin (Ink splits them)

deng-diana:
Completes #51. On a real **default Terminal.app** (claude-cues 2.1.158), #51's event-level synth doesn't fire: **Ink splits the `\x1b\x1b[A` chord into a standalone escape + a plain arrow before any handler sees it**, so the arrow loses the double-ESC prefix the synth gates on. Probe evidence: `{key:"escape",seq:"\x1b"}` then `{key:"up",seq:"\x1b[A",synthFired:false}`, same millisecond.

## Fix — one layer earlier (no Ink fork, no cli.js patch)
Rewrite the raw stdin chunk before Ink parses it: `\x1b\x1b[A/B/C/D` → `\x1b[1;7A/B/C/D` (the modifier-CSI Ghostty/iTerm2 already send, which Ink decodes to `{ctrl,alt}`). Ink/CC pull stdin via `'readable'` + `read()` with `setEncoding('utf8')`, so the installer wraps `read()` (utf8 strings) + `emit('data')`, installed once from the CC `boot()` seam.

- `packages/opencues-runtime/src/modules/mac-keyboard.ts` — pure `rewriteMacDoubleEscArrows` (Buffer) + `…String` (utf8) + `installMacDoubleEscStdinRewrite`.
- `adapters/cc/v2.1/boot.ts` — one install call.

## Safety
- **Escape-safe** by the contiguous-byte invariant: the chord's 4 bytes land in one read buffer; a real Escape arrives alone → matching only within one buffer can't swallow Escape. No state, no timer, no Escape latency.
- **Windows/Linux: strict no-op** — darwin-gated; the installer returns early before wrapping stdin off macOS.
- Degrades to prior behaviour on split-chunk transports (tmux/ssh).

## Verification
- Real default Terminal.app: Ctrl+Option+arrow now arrives ctrl+alt; up/down cycle confirmed by the tester.
- 61 unit tests incl. a real-`Readable` scenario test mirroring Ink's `read()`-loop + `setEncoding('utf8')`. Reviewed by two subagents (APPROVE).
- Docs: `adapters/cc/REPAIR.md` §14 + integration `CLAUDE.md` #4.

Follow-up (not this PR): gemini-cli has the same Ink-split issue and is fixable by the same installer in its bootstrap.


wkasekende:
Superseded by #57 — rebased onto master with conflict resolutions (mac-keyboard.ts kept both `buildOpenTuiModifiers` and stdin-rewrite functions, version bumped to 0.1.12). Your two commits + authorship preserved in #57. Thank you for the device-level verification and the byte-stream evidence!

PR #77, opened 2026-06-08
perf(blank-fill): per-blank result cache skips repeat spawns within TTL

wkasekende:
## Summary

- Per-blank result cache in BlankFill. Repeat invocations with identical args within TTL skip the spawn entirely.
- Saves **~150ms** per repeat WSL spawn (volume/brightness) and **~500ms** per repeat network call (weather/stocks/HN/crypto/claude-status). Visible on cycle-then-cycle-back + backspace+retype recovery.
- New `blankCacheTtlMs` frontmatter field (default 2000ms, 0 disables).
- **Intended merge order: after #74 + #76.** Version bumps assume that sequence — `@opencues/core` 0.3.1 → 0.3.2 (was 0.3.0 → 0.3.1 in #76), `@opencues/runtime` 0.2.4 → 0.2.5 (was 0.2.3 → 0.2.4 in #76). Rebase as needed if you merge in a different order.

### `@opencues/core` (0.3.1 → 0.3.2)
- New `blankCacheTtlMs?: number` on `BlankConfig` + frontmatter struct + parser case. Strict integer parse; negatives + non-numeric drop silently.

### `@opencues/runtime` (0.2.4 → 0.2.5)
- `BlankFill._resultCache: Map<string, {output, fetchedAt, ttlMs}>` with LRU semantics (insertion-order Map + cap 32 entries).
- Cache-hit path: emits `blank.invoked` with `cacheHit: true` for observability, bumps LRU recency, clears the pending-dedup entry, calls `applyAsyncFill` directly.
- Cache write: only on `exitCode === 0 && stdout` — failures un-cached so retries are cheap.
- Default `DEFAULT_CACHE_TTL_MS = 2000`. `blankCacheTtlMs: 0` disables for that blank.

## Test plan

- [x] 4 new regression tests (`BlankFill result cache` block in `blank-fill.test.ts`):
  - identical-args repeat → cache hit (no second spawn)
  - past-TTL → cache miss (spawn fires again)
  - `blankCacheTtlMs: 0` → cache disabled (every call spawns)
  - failure (exit≠0) → NOT cached → next call spawns
- [x] Tests use a `reArmAndPush` helper that mirrors the real user keystroke flow (clear buffer + re-type `_` so the explicit-`_` gate arms).
- [x] Full runtime suite: 1592 / 1592 pass.
- [x] Full core suite: 158 / 158 pass.
- [ ] Manual smoke on `claude-cues` post-merge — type `volume _` twice in 2s; second call should be instant.

## Why this PR doesn't touch the shipped `defaults/blanks/*.md`

Every blank gets the 2000ms default which is safe for action blanks. Ambient blanks (stocks / weather) would benefit from a longer TTL but tuning them is a separate trade-off (correctness window vs cache hit rate) that should go through its own bench. Authors can opt any blank in via `blankCacheTtlMs: <ms>` in BLANK.md today.


wkasekende:
Superseded by re-PR with rebased commit on master (clean single-commit history).

PR #78, opened 2026-06-08
fix(resolver): parallel mode enforces claim-then-bail semantics post-dispatch

wkasekende:
## Summary

- The runtime resolver always passes `parallel: true`. Pre-fix, the parallel branch couldn't enforce `consumedBlankSlots` claims either at dispatch (all sources started simultaneously, none saw each other's claims) or after (the merge didn't filter overlapping results).
- The user-visible symptom: TransformBlank claims a `_` slot but emits no rewrite (claim-and-bail), FluidBlank in the same batch produces a stray lookup answer for the same `_`, FluidBlank's answer substitutes and "vandalises" the user's compose intent.
- Post-fix: post-dispatch reconciliation walks sources in priority-desc order, accumulates claims as it goes, and filters each source's results against accumulated claims (excluding the source's own claim).
- **Correctness fix, not a perf change.** Same parallel dispatch latency; reconciliation is in-memory post-processing.
- **Intended merge order: after #74, #76, #77.** Version bump assumes that sequence — `@opencues/core` 0.3.2 → 0.3.3.

### `@opencues/core` (0.3.2 → 0.3.3)
- `CueResolver.resolve`'s parallel branch:
  1. After `Promise.all`, walk sources in priority-desc order (constructor already sorts that way).
  2. For each source: filter its results against the accumulated `consumedBlankSlots` set.
  3. Accumulate the source's `consumedBlankSlots` from its result.
  4. ALSO accumulate the source's produced `wordIndex` into the same set — a higher-priority source's content claim suppresses lower-priority results at the same index (closes the same-priority tiebreak corner).

## Test plan

- [x] 4 new regression tests in `packages/opencues-core/src/resolver.test.ts` under a new describe block:
  1. Lower-priority result dropped on higher-priority claim-and-bail
  2. Lower-priority result dropped on higher-priority content claim
  3. Different-`wordIndex` results survive (only overlapping ones filtered)
  4. A source's own `consumedBlankSlots` does NOT filter its own results
- [x] Full core suite: 158 vitest + 801 node:test pass.
- [x] Full runtime suite: 1588 / 1588 pass.

## Follow-up not addressed by this PR

Lower-priority sources still PAY for the LLM call even when their result is dropped at reconciliation. In-batch sibling cancellation (abort other sources' calls when a higher-priority source claims a shared slot) is a separate piece of work — #76 covers the supersede-by-new-keystroke case but not the intra-batch race.


wkasekende:
Superseded by re-PR with rebased commit on master (clean single-commit history after #74/#84/#85 squash-merged).

PR #80, opened 2026-06-08
perf(config-loader): parallelise every independent fs read in _loadOnce

wkasekende:
## Summary

- `ConfigLoader._loadOnce` ran fs reads in serialised waves; every read was independent of every other wave's, but `await` boundaries forced order.
- Post-fix: every independent read fans out under one top-level `Promise.all`; the two for-loops inside `_discoverFolders` become `Promise.all` over entries and scopes.
- Downstream merge / fold logic unchanged. Same priority semantics, same fold-low-to-high overlay rule.
- **No API surface changes.** Callers see the same `load()` Promise resolving with the same result shape, just faster on cold reads.
- **Intended merge order: independent of other PRs.** Runtime version bump assumes #74/#76/#77/#79 land first (next number = 0.2.7).

### `@opencues/runtime` (0.2.6 → 0.2.7)
- `_loadOnce` top-level Promise.all hoists 4 categories of fs work (settings, identity, master batch, folder discovery) into one parallel wave.
- `_discoverFolders`:
  - Per-entry for-loop (`for (const e of entries)`) → `Promise.all(entries.map(...))`.
  - Per-scope for-loop (`['cues', 'blanks', 'auditors']`) → `Promise.all(...)`.

## Test plan

- [x] Existing config-loader suite: 33 / 33 pass.
- [x] Full runtime suite: 1588 / 1588 pass.
- No new tests — parallelisation is an implementation detail that preserves output semantics. A "parallelism" test would essentially time the load and is too host-dependent to pin reliably.

## Estimated win

50-200ms per cold `_loadOnce` on WSL / mounted FS. Invisible on a fast SSD. The 2s `maybeReload` debounce + 5s background poll mean this fires at most a few times per second of UI activity; the cumulative saving over an hour-long session is small but visible during cold-startup spikes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 > **Scope of this section**: only changes tied to an actual package version bump are listed. The project shipped many other features and fixes since 0.1.0 (sentence cues, auditors, agent-rewrite, ambient/user context, etc.) without bumping versions at the time — those landed in source but aren't formally versioned, so they're tracked in git, not here. From now on, the rule in `docs/architecture/versioning.md` § Discipline keeps changelog entries and version bumps shipping together.
 
+### Changed — ConfigLoader parallelises every independent fs read on each `_loadOnce`
+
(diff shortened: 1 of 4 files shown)


wkasekende:
Superseded by re-PR with rebased commit on master.

PR #81, opened 2026-06-08
fix(blank-fill): staleness check accepts loading-frame chars (closes latent silent-drop bug)

wkasekende:
## Summary

- Closes the latent silent-drop bug introduced by `0097d65` (2026-05-28 refcount commit). BlankFill's `applyAsyncFill` staleness check rejected any non-`_` char in the slot — including the loading animator's frame chars when the resolver outlives BlankFill's release. PR #74 made the typical race-window so small the bug rarely fires in production, but the structural latent bug remains for any code path where the resolver outlives BlankFill's release.
- New `BlankLoadingAnimator.isOurSlotChar(wordIndex, char)` predicate lets BlankFill distinguish "still our slot, just currently painted as a loading frame" from "user typed over the slot".
- **No latency change** — purely a correctness fix.
- **Intended merge order: independent of other PRs.** Runtime version bump assumes #74/#76/#77/#79/#80 land first (next number = 0.2.8).

### `@opencues/runtime` (0.2.7 → 0.2.8)
- New public predicate `BlankLoadingAnimator.isOurSlotChar(wordIndex, char): boolean` returns true for the literal `_` or any of the slot's currently-active frame characters (per-slot frames so user-supplied custom frames in `blank-loading-frames` are also recognised).
- `BlankFill.applyAsyncFill` consults the predicate via `this._loading?.isOurSlotChar(slot.index, target.word) ?? false` — staleness now means "user typed a real character over our slot", not "the animator painted a frame here".

## Test plan

- [x] 5 new regression tests in `packages/opencues-runtime/src/modules/blank-loading.test.ts` under a new describe block:
  1. `_` recognised even for unknown slot
  2. Bounce-frame char in active slot recognised
  3. Non-frame, non-`_` rejected (real user-typed-over case)
  4. Frame-char query against inactive slot rejected
  5. Custom-mode user-supplied frames recognised
- [x] Full runtime suite: 1593 / 1593 pass.

## Follow-up — Item #5 in the perf audit (animator → render-directive overlay)

The larger refactor that would structurally remove this entire bug class by NOT painting frame chars into the buffer in the first place — emit a render directive instead and let the host overlay the glyph at render time. Eliminates the reconciler churn during spinner too. Defer to a future PR; this one closes the immediate latent silent-drop without the full refactor surface.


wkasekende:
Superseded by re-PR with rebased commit on master.

PR #83, opened 2026-06-08
feat(linkedin): Quill write path for LinkedIn share composer

wkasekende:
## Summary

LinkedIn migrated their share composer from ProseMirror to Quill.js. Symptoms on master + PR #82:
- TransformBlank substitutes (e.g. \`draft an email _\`) briefly land then **disappear** — Quill's MutationObserver reverts external DOM mutations.
- LinkedIn does NOT expose \`__quill\` on any standard probe location (verified via DevTools), so any path relying on the Quill API instance can't fire on LinkedIn.

### 1. PREFERRED: \`__quill.clipboard.dangerouslyPasteHTML(html, 'user')\`

Managed-shape HTML (\`<p>\` per paragraph, \`<br>\` for soft breaks — inherits the new paragraph emission from PR #82). Writes through Quill's HTML paste pipeline, Delta model owns the DOM. Single-margin paragraph rendering, no double-spacing.

### 2. SECONDARY: \`__quill.setText(condensed, 'user')\`

\`condensed = text.replace(/\n\n+/g, '\n')\`. Used when clipboard module isn't exposed but \`setText\` is. \`\n+\` collapse avoids empty paragraph blocks in Quill's Delta.

### 3. FALLBACK: selectAll via Range API + \`execCommand('insertText')\`

\`\`\`ts
const range = document.createRange();
range.selectNodeContents(target);
window.getSelection().addRange(range);
document.execCommand('insertText', false, condensed);
\`\`\`

This is what actually fires on LinkedIn since \`__quill\` isn't reachable. The prior synthetic-paste fallback (Ctrl+A keydown + ClipboardEvent with text/html) had Quill rejecting the untrusted paste — substitute reverted to clean state. The Range-API selection + insertText path is observed working: Quill's \`beforeinput\` handler reads the \`insertReplacementText\` inputType and processes it through its Delta model. Same code path real typing uses, so the model accepts the write instead of reverting.

## Why stacked on PR #82

PR #82 already added:
- Capture-phase input listener (Quill stops propagation in bubble too)
- Runtime gate fallback (LinkedIn share composer is also a focus-trap modal → blank dispatch was silently blocked)
- Managed-editor paragraph emission (Quill inherits the new \`<p>\` per paragraph + \`<br>\` soft-break shape via path 1 above)

Without PR #82, LinkedIn would never even fire the blank — let alone reach this PR's write path. Without this PR, the blank fires but the substitute disappears. Both are required for LinkedIn to work end-to-end.

This PR targets PR #82's branch — should land after #82 or be retargeted to master if #82 lands first.

## Tests (4 new)

\`replace-all-text-undo.test.ts\`:
1. Preferred path → \`clipboard.dangerouslyPasteHTML\` with managed-shape HTML; no Ctrl+A.
2. Secondary path → \`setText\` with \`\n\n+\`-condensed text when clipboard module unavailable.
3. Fallback → \`execCommand('insertText')\` with condensed text; no Ctrl+A, no synthetic paste.
4. \`dangerouslyPasteHTML\` error → falls through to \`setText\` cleanly.

## Verified end-to-end on LinkedIn share composer

- \`volume _\` → already worked (small in-place \`diffWriteText\` splice, Quill absorbs it).
- \`draft an email _\` → substitute now lands and STAYS (was reverting before this PR).
- Paragraph rendering: single-margin gaps from the new managed-editor emission.

## Test plan

- [x] LinkedIn share composer: \`draft an email _\` substitutes and stays
- [x] LinkedIn share composer: \`volume _\` still works (regression check)
- [ ] Regression smoke on Reddit (Lexical), ChatGPT/claude.ai (PM), Twitter (Draft.js), Gmail (generic CE) — the Quill branch is new and gated on \`.ql-editor\` selector so other engines shouldn't hit it, but worth confirming

### DIFF
--- a/integrations/chrome/CLAUDE.md
+++ b/integrations/chrome/CLAUDE.md
@@ -498,8 +498,8 @@ real failures the user should see immediately.
 | YouTube comments | generic contenteditable | same | 1 |
 | Reddit | Lexical | `editor.update(() => { root.clear(); insertNodes(...) })` — one transaction, OR fallback: Ctrl+A keydown + synthetic paste | 1 |
 | Twitter/X | Draft.js | Ctrl+A keydown + synthetic paste with `text/plain` | 1 |
-| LinkedIn | ProseMirror | `execCommand('insertHTML', false, '<p>...</p>')` over select-all | 1 |
-| ChatGPT | ProseMirror | same | 1 |
+| LinkedIn (share composer) | Quill.js | `editor.setText(text, 'user')` via `__quill` on `.ql-container` — single Delta op. Fallback: Ctrl+A keydown + synthetic paste with `text/plain` | 1 |
+| ChatGPT | ProseMirror | `execCommand('insertHTML', false, '<p>...</p>')` over select-all | 1 |
 | claude.ai | ProseMirror | same | 1 |
 | Luma | TipTap | Ctrl+A keydown + synthetic paste with `<p>` HTML — historically the "outlier", retained as its own branch but Luma's TipTap also accepts the default managed `insertHTML` path (May 2026 verified). The keyboard-sim path is left in for now until a broader TipTap regression sweep clears collapsing the branch. | 1 |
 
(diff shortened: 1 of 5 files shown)


wkasekende:
Superseded by #91 — same work landed after rebase onto master.

PR #106, opened 2026-06-09
sec(user-blanks): F1 stopgap — load-time warn + honest audit (INFOSEC F1)

wkasekende:
## Summary

**The critical finding.** Node's `vm.runInContext` is not a security boundary against adversarial JS — `Promise.constructor('return process')()` reaches the host realm, then `process.env` exposes every API key and `child_process.execSync` runs arbitrary commands. Live-confirmed June 2026 against the shipped sandbox composition.

**This PR is the stopgap, not the fix.** The structural fix (migrate to `isolated-vm` or out-of-process Node with `--experimental-permission`) is ~1-2 weeks of integration work and warrants a separate PR + benchmark sweep + chrome-bundle native-module consideration. Tracked as an Open follow-up in `security-audit.md`.

Stopgap changes:
- **Header comment in `node-loader.ts` rewritten** to reflect actual posture. The prior comment falsely claimed primitive wrappers couldn't reach the host realm. New comment names the constructor-chain escape explicitly.
- **One-time loud warn** on first load of each blank path. `console.warn` + adapter log. Dedup'd per-process so steady-state isn't spammed. Mentions "INFOSEC F1", "FULL host privileges", "constructor-chain", and points at `opencues review`.
- **`security-audit.md` row #2** flipped 🟢 → 🟡 with honest residual-risk description.
- **`security-audit.md` § Open follow-ups** gains an entry naming `isolated-vm` vs out-of-process as the candidate fixes + the integration cost.

INFOSEC findings doc, finding F1.

## Test plan

- [x] 5 new tests in `node-loader.f1-warn.test.ts` cover: first load warns, dedup works, different paths warn independently, `_resetF1WarnCache` works, warn flows through adapter log
- [x] All 24 node-loader tests + 52 user-blank tests still pass
- [x] `lint-version-bump.sh`, `lint-legacy-names.sh` clean
- [x] Runtime builds cleanly

## Notes — what this PR does NOT do

- Does NOT replace `vm.runInContext` with a real isolate. That's the real fix, tracked separately.
- Does NOT refuse to load `impl: ./blank.js` packs. Backwards-incompatible — would break every existing custom-JS install.
- Does NOT add a per-pack consent prompt at install time. `opencues review` (F5) is the install-time gate today; its denylist now refuses the most common escape patterns. A consent prompt would belong with the registry work tracked in the pre-registry follow-ups.

## Migration path for the real fix

When `isolated-vm` lands:
1. Add `isolated-vm` as a runtime dep (native module — needs build-time consideration for the chrome bundle path).
2. Replace `vm.createContext` + `vm.runInContext` with `IsolatedVM.createSnapshot` + `Isolate.createContext`.
3. Drop the F1 stopgap warn (the warn copy then becomes inaccurate).
4. Re-benchmark per-invocation cost — keep cache-hit case sub-ms.
5. Flip security-audit.md row #2 amber → green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 > **Scope of this section**: only changes tied to an actual package version bump are listed. The project shipped many other features and fixes since 0.1.0 (sentence cues, auditors, agent-rewrite, ambient/user context, etc.) without bumping versions at the time — those landed in source but aren't formally versioned, so they're tracked in git, not here. From now on, the rule in `docs/architecture/versioning.md` § Discipline keeps changelog entries and version bumps shipping together.
 
+### Security — F1 stopgap: load-time warning + honest audit + tracked follow-up (INFOSEC F1)
+
+The critical finding from the static + dynamic security review: Node's `vm.runInContext` is NOT a security boundary against adversarial JS. `Promise.constructor('return process')()` reaches the host realm, then `process.env` exposes every API key and `child_process.execSync` runs arbitrary commands. Live-confirmed June 2026 against the shipped sandbox composition. The structural fix (migrate to `isolated-vm` or out-of-process Node with `--experimental-permission`) is ~1-2 weeks of integration work and a separate PR — this PR is the stopgap.
+
+- **`@opencues/runtime` (0.2.8 → 0.2.9)** — `node-loader.ts`:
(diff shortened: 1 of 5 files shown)


wkasekende:
Superseded by #108 (the isolated-vm real fix). The stopgap warn this PR shipped is no longer load-bearing once #108 lands — node-loader.ts gets rewritten to use isolated-vm and the F1 escape becomes structurally impossible, so the warn becomes inaccurate.

PR #121, opened 2026-06-11
fix(runtime): BlankFill skips keyword match inside substituted DynDef span

wkasekende:
## Summary

`BlankFill` looped on substituted output whose text contained a registered keyword. Canonical reproducer + symptom you reported:

```
type: nvda _ + apple _ = _
got:  Nvidia NVDA: $200.42 + apple • = •     (apple + = stuck loading forever)
```

`nvda _` substitutes to `Nvidia NVDA: $200.42`. The substituted span contains `nvidia` (one of the stocks blank's keywords). On the next text-change BlankFill scanned the new buffer, matched `nvidia` against the next `_` further along, and re-fired the substitute. Buffer kept looping; sibling blanks (`+ apple _`, `= _`) never got a stable scan to fire against, so they sat in their loading frames forever.

Event-stream evidence on the pre-fix run:

```
blank.substituted nvda → NVDA: $200.42  (ts 218140)
blank.substituted apple → AAPL: $291.58 (ts 316164)
blank.substituted nvda → NVDA: $200.42  (ts 339598)  ← repeat
blank.substituted nvda → NVDA: $200.42  (ts 415878)  ← repeat
blank.substituted nvda → NVDA: $200.42  (ts 576113)  ← still firing
```

## Fix

`BlankFill.matchKeyword` now consults `DynDefs.findSpanContaining` for each candidate keyword word index and skips matches that fall inside an already-substituted multi-word span. `findSpanContaining` only returns multi-word spans (≥2 words), so single-word substitutions stay matchable — that's the mirror of how the broader runtime treats single-word DynDefs as the original-keyword-is-unchanged case.

## Test plan

Two regression tests pinned in `blank-fill.test.ts`:

- `skips keyword match inside an already-substituted multi-word span` — simulates the post-`nvda _` substitute state (DynDef span 0-2 with `Nvidia NVDA: $200.42`); scans `Nvidia NVDA: $200.42 + apple _`; expects only the apple slot to match.
- `still matches keyword OUTSIDE substituted spans` — belt-and-braces; DynDefs has a single-word substitute elsewhere, the apple keyword outside any span still matches.

All 1642 runtime tests pass.

## Known follow-up (separate PR)

After this fix, the harness still shows `apple •` and `= •` stuck. Event-stream evidence:

```
blank.invoked apple   (ts 910189)
blank.substituted nvda  (ts 910371)
```

apple's invoke fires but its substitute never lands. This is a SEPARATE bug — concurrent invokes coordinating with span writes — and is being tracked as a follow-up. This PR ships the loop fix because (1) it eliminates the most visible symptom, (2) the fix is structurally tight + well-tested, and (3) the concurrent-invoke bug is harder to reproduce reliably and needs its own focused investigation.

## Version

`@opencues/runtime` 0.3.2 → 0.3.3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 > **Scope of this section**: only changes tied to an actual package version bump are listed. The project shipped many other features and fixes since 0.1.0 (sentence cues, auditors, agent-rewrite, ambient/user context, etc.) without bumping versions at the time — those landed in source but aren't formally versioned, so they're tracked in git, not here. From now on, the rule in `docs/architecture/versioning.md` § Discipline keeps changelog entries and version bumps shipping together.
 
+### Fix — BlankFill stops looping on substituted output that contains a registered keyword
+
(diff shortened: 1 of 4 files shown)


wkasekende:
Superseded — the BlankFill DynDef-span guard commit was absorbed into PR #123 (merged at 6b86b57). Closing this branch as redundant.

PR #157, opened 2026-06-16
fix(runtime): resolver explicit-_ gate uses e.previousText, not stale _lastInputText

wkasekende:
## Summary

Stacked-blank scenarios silently broke after #155 / #156 made BlankFill substitutes work properly. Repro: user types \`config _\` → BlankFill substitutes to \`fluid-blank-mode on\` → user appends \` build a website _\` → **no TransformBlank, no FluidBlank, no ConfigIntent**. The appended \`_\` got masked from every blank source.

## Root cause

\`Resolver._lastInputText\` updates only on \`source: 'user'\` events (early-return at \`resolver.ts:843\`). BlankFill's satellite substitute fires \`setText\` with \`source: 'runtime'\`, so \`_lastInputText\` stays at the pre-substitute text. The next user keystroke then compares against a stale baseline:

| | Stale \`_lastInputText\` (bug) | Actual prior buffer \`e.previousText\` |
|---|---|---|
| prev | \`config _\` (ends with \`_\`) | \`fluid-blank-mode on\` (no \`_\`) |
| \`blankJustTyped\` | false | true |
| underscore count | 1 == 1, not fresh | 0 → 1, fresh |
| Result | masked, \`allowBlanks=false\`, no-op | fires |

## Fix

Gate now reads \`e.previousText\` (the adapter's actual prior buffer state, updated regardless of source) for trigger-detection + freshness comparisons. \`_lastInputText\` keeps its same-text-dedupe role at the early-return.

## Why the test suite missed it

- 95% of resolver tests use single-\`_\` buffers — never reach a state where \`_lastInputText\` is stale AND new text has a \`_\`.
- The few stacked-blank tests don't go through a \`source: 'runtime'\` write in between.
- The explicit-\`_\` gate was added without a scenario that paired it with a BlankFill substitute first.

The new scenario test exercises the full three-step sequence (\`config _\` → satellite substitute via \`adapter.setText\` → user appends with \`_\`). It catches this and would have caught it pre-#155/#156.

## Test plan

- [x] New scenario test \`user-typed _ after a runtime substitute (stale-baseline regression)\`
- [x] Verified the test FAILS without the fix and PASSES with it
- [x] \`pnpm --filter @opencues/runtime exec vitest run\` — 1673/1673 green
- [x] Verified in the agentic harness: \`transform-blank.started\` + \`fluid-blank.started\` events now fire for stacked \`_\` (vs. silent no-op pre-fix)
- [x] CHANGELOG.md + \`@opencues/runtime\` 0.3.16 → 0.3.17 bumped

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 > **Scope of this section**: only changes tied to an actual package version bump are listed. The project shipped many other features and fixes since 0.1.0 (sentence cues, auditors, agent-rewrite, ambient/user context, etc.) without bumping versions at the time — those landed in source but aren't formally versioned, so they're tracked in git, not here. From now on, the rule in `docs/architecture/versioning.md` § Discipline keeps changelog entries and version bumps shipping together.
 
+### Fix — shape gate tries both keyword-window AND full-prefix slices
+
+Follow-up to #155. The shape-anchor fix sliced from the keyword's leading word so stacked invocations (`<prior-span> + nvda _`) could match. That broke prompt-improver's `^(.+?)\s+improve\s+prompt\s*_$` trigger-at-end shape: `(.+?)` had nothing to capture in the keyword-only slice, so `build a website improve prompt _` silently no-op'd.
+
+Two shape conventions coexist now:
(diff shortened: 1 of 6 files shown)


wkasekende:
Closing unmerged. The shape system from PR #146 has been reverted on master (commits 2c9aa2a, 44969ba, 878aaa7) and preserved on `dev/shape-system` for future exploration. The fixes in this PR + #156 + #155 are all on `dev/shape-system` if/when we re-explore the system on a dedicated branch.

PR #162, opened 2026-06-18
refactor: unify three semantic-_ sources onto SPAN splice + extract VariantCache + fix CC harness path

wkasekende:
## Summary

Architectural unification thread the user kicked off: *less concepts in the system, more reliance on the model*. Four commits stacked, each independently meaningful:

### 1. SPAN-splice unification (`27ba00b` · core 0.3.27)
All three semantic-\`_\` sources (FluidBlank, ConfigIntent, FUSED TransformBlank) now share **one** substitute mechanism — LLM emits a verbatim \`SPAN\` substring + slice fill, runtime locates via \`findSpanCharRange\`, splices. ConfigIntent's old \`[0, text.length)\` wipe-everything is gone; FUSED TransformBlank's old whole-buffer + \`threeWayMerge\` path retires. \`word-diff.ts\` stays load-bearing only for AgentRewrite (where in-flight typing protection genuinely needs it). Bench: prod-fused on cerebras gpt-oss-120b at **186/231** (inside the historical 186–193 noise band).

### 2. \`VariantCache<T>\` extraction (\`c5f9f63\` · core 0.3.28)
Three sources each had a private \`static _variantPool = new Map<...>()\` with identical LRU + cycling logic. Extracted as one generic primitive (\`packages/opencues-core/src/variant-cache.ts\`); -120 lines of duplicated state-machine code, +1 documented class + 10 unit tests. Public API preserved (\`variantPoolSize()\` / \`resetVariantPoolForTest()\`) so no source-test edits required.

### 3. event-bridge \`text:\` inject order (\`8686ef5\` · runtime 0.3.19)
Surfaced after fixing the agentic harness's CC PTY launcher: scenarios 100/102/103 timed out on CC. Root cause: bridge called \`adapter.setText(decoded)\` BEFORE \`notifyTextChange(...,'user')\`. On hosts whose setText eagerly updates \`lastSeenText\` (CC v2.1 does — to drift-track React re-render echoes), \`previousText\` then equalled \`text\` and the Resolver's explicit-\`_\` gate saw no diff → blank trigger silently masked. Fix: notify first, set second. Shell + Gemini bind setText directly to the host so they're unaffected by the order.

### 4. CC \`onCursorChange\` wiring + \`cursor:\` order fix (\`6fca0e1\` · runtime 0.3.19)
CC's adapter had no \`onCursorChange\` at all — \`cursorHandlers\`, \`registerCursorChangeHandler\`, \`notifyCursorChange\` all absent. Scenario 30 (cursor-navigate mode → plain Right arrow auto-activates highlight) silently no-op'd because \`adapter.onCursorChange\` was undefined. Wired the path end-to-end + mirrored the prev-stale fix to the bridge's \`cursor:\` command.

## Verified

| Surface | Result |
|---|---|
| Core unit tests | 49/49 pass |
| Runtime scenario tests (\`transform-blank.scenarios.test.ts\`) | 26/26 pass |
| Agentic shell suite | 61/63 (96%) — 2 pre-existing flakes |
| Agentic CC suite | **72/77 (94%)** — 5 pre-existing flakes (model-override, runaway-loop japanese, fluid-config sequential, blank-fill cache TTL, pipeline events) |
| transform-blank prod-fused bench (cerebras) | 186/231 (inside historical noise band) |
| Scenarios 100/102/103 (prior-content preservation) | 9/9 on CC, 12/12 on shell across stability runs |

CC's pass rate moved from **64/77 → 72/77** purely from infra fixes (CC suite couldn't even hit these code paths before).

## Test plan

- [ ] Re-run agentic suite on shell + CC after merge to verify no drift from main
- [ ] Manually exercise: type \`hii world. voice mode off _\` on each host — prior \`hii world.\` preserved
- [ ] Manually exercise: stacked transforms (\`make wilfred bold _\` → \`make it caps _\`) on CC — both apply, markdown survives

## What's next (not in this PR)

AgentRewrite is the only remaining production user of \`threeWayMerge\`. Decision point on whether to migrate it to deltas-based output (changes LLM contract, requires bench re-run) or keep whole-buffer + diff (current good path). Tracked separately — wants empirical bench data first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 > **Scope of this section**: only changes tied to an actual package version bump are listed. The project shipped many other features and fixes since 0.1.0 (sentence cues, auditors, agent-rewrite, ambient/user context, etc.) without bumping versions at the time — those landed in source but aren't formally versioned, so they're tracked in git, not here. From now on, the rule in `docs/architecture/versioning.md` § Discipline keeps changelog entries and version bumps shipping together.
 
+### Refactor — unify ConfigIntent + FUSED TransformBlank onto FluidBlank's SPAN splice (core 0.3.27)
+
+All three semantic-`_` sources (FluidBlank, ConfigIntent, FUSED TransformBlank) now share **one** substitute mechanism: the LLM emits a verbatim `SPAN` substring + a slice fill, the runtime locates the span via `findSpanCharRange` (shared helper exported from `fluid-blank-source.ts`), and splices the fill into that range. Anything outside the SPAN is preserved verbatim — fixing a real bug where typed prior content was being wiped.
+
+Concretely:
+
+- **ConfigIntent** (`config-intent-source.ts`): classifier prompt now emits `SPAN:` alongside `INTENT/SETTING/VALUE`. The result's `spanStart`/`spanEnd` come from `findSpanCharRange(verdict.summon, context.text)` instead of the old `[0, text.length)` wipe-everything. `"hii world. voice mode off _"` now leaves `"hii world."` in place and only wipes `"voice mode off _"`.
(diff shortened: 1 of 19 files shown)


wkasekende:
Closing — superseded by the single-fused-pipeline direction.

#162 unified ConfigIntent + fused TransformBlank onto SPAN-splice and kept the 3-pass path. We instead went the other way: **retired 3-pass entirely** and collapsed TransformBlank to a single fused call (#196), after benchmarking showed groq fused at parity with 3-pass and ~35% faster (EXPERIMENTS.md § Experiment 10). The SPAN-splice mechanism this PR introduced conflicts with that whole-buffer FULL_REWRITE → three-way-merge design, and the branch also predates the model-override removal (#165), so it carried 15+ references to the deleted `model-aliases` module.

What's been salvaged from here already:
- The clean runtime fixes (event-bridge inject order, CC cursor wiring, fluid-blank.completed race) → re-landed in #194.
- The genuine capability gaps it aimed at (list-ification, cursor/deictic) → addressed directly in the fused prompt (#197 STRUCTURE rule, #198 [CURSOR] anchor), evidence-gated by bench + agentic.

The VariantCache<T> extraction idea remains the one un-salvaged piece; can revisit as a standalone refactor if cache-dedup becomes worth it. Branch deleted.

PR #166, opened 2026-06-19
fix(core,runtime): data-loss hardening — gate `seed`, harden FluidBlank splice, kill the config-intent nuke

wkasekende:
## Why
Follow-up to the `prediction`-param gate — an audit found the **same ungated-param shape** elsewhere, plus a splice-site gap.

## Fixes (all structural, no heuristics)
- **`seed` gated to native providers** (cerebras / groq / openai). Every source hardcodes `seed: 42`; it was emitted to *all* OpenAI-shape providers — the exact twin of the `prediction` 400 bug. On an openrouter→anthropic route it 400s. Now never sent to a pass-through gateway. Determinism (variant cache) only pins groq/cerebras, so nothing is lost.
- **AgentRewrite legacy inline path gates `reasoning_effort`** by model-name (it bypasses `buildOpenAIBody` entirely; `reasoning_effort: 'low'` 400s on Groq llama-*).
- **FluidBlank WIPE splice re-validates the live buffer** over the splice range (mirrors the guards ConfigIntent + TransformBlank already have) — until now the destructive WIPE relied solely on the source's parser-bounded span.

A `<10%` fused-collapse guard was **considered and rejected** — a length ratio can't tell a hallucinated collapse from a legitimate `summarize _`, and would silently drop valid transforms.

> Interim point-gates. The follow-up is a **provider capability model** so "forgot to gate a provider-specific param" becomes structurally impossible.

## Tests
Core 991 + 177, runtime 1647 — 0 failures. New `seed`-gate unit test (openrouter omits, cerebras/groq/openai keep).

**Agentic suite (67 scenarios, opencode):** no new failures introduced — the 5 failing scenarios fail **identically on master** (verified by reinstalling master + re-running), so they're pre-existing and out of scope here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed — gate provider-specific request params + harden the FluidBlank splice site (core 0.3.29 / runtime 0.3.21)
+
+Follow-up to the `prediction`-param removal: an audit found the **same ungated-param shape** elsewhere, plus a splice-site gap.
+
+- **`seed` gated to providers that natively support it** (cerebras / groq / openai). Every source hardcodes `seed: 42` and it was emitted to *all* OpenAI-shape providers — the exact structural twin of the `prediction` bug. On an openrouter→anthropic route it would 400 (anthropic has no `seed`). Now never sent to a pass-through gateway. The variant-cache determinism that relies on seed only pins groq/cerebras, so the gate loses nothing.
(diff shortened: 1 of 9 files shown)


wkasekende:
Expanded scope after the agentic-suite investigation: added a third data-loss fix — **ConfigIntent no longer nukes the whole buffer**. It hardcoded `spanStart: 0` and spliced the settings pair over `[0, len)`, so `hii world. voice mode off _` lost "hii world.". Now wipes only from the trailing summon phrase (`summonPhraseStart`). Pinned by unit tests + agentic scenario **102 (now 15/15, was the documented bug-repro)**.

Sweep result: ConfigIntent was the **only true lazy-nuke** (short pair spliced over the whole buffer, no merge). TransformBlank's whole-buffer spans go through the three-way merge (content-preserving); the residual `#103` (LLM occasionally drops the prefix in FULL_REWRITE) is a separate SPAN-field architectural question, not a nuke.

PR #203, opened 2026-06-24
feat(blank-intent): typed-SET — volume 30 _ / brightness 70 _ actually set

wkasekende:
## What

BlankIntent (PR #201, merged) already **extracted** the action+value from an invocation — `volume 30 _` → `INVOKE volume set 30` — but the runtime **discarded** it and always ran `get`. So typing a value silently showed the *current* value instead of setting:

```
BlankIntent: INVOKE volume set 30        ← gate extracts "set 30"
BlankFill: invoke volume get volume      ← runtime runs GET, ignoring 30
substituting "volume _" → "54%"          ← shows current, not 30
```

This threads the verdict into the dispatch so typed-number SET actually works.

## How

- The gate contract widens from `'invoke' | 'cede'` to a `BlankIntentDecision` (`{ verdict, action, value }`). The classifier already produced this; the boot wrappers just stop collapsing it.
- In `maybeRunScripts`: an `action: 'set'` + **numeric** value on a **settable** blank (one with `blankStep` → volume/brightness) runs `set <value>` (`runBlankSet`, mirroring Cycling's `invokeOrSpawn('set', …)`), then falls through to the existing `get` to read the **clamped** value back and splice it.
- Everything else is unchanged and safe:
  - **Non-settable** blanks (weather/stocks/dictionary/…) ignore a stray `set` → plain `get` (their value is a lookup query, never a write).
  - **Non-numeric** values degrade to `get` (defensive).
  - Both wrappers **degrade to a get-invoke** on error / no-key (never a spurious set).
  - Consent unchanged — `set` only reaches a blank whose keyword the user typed; it's bounded 0–100, local, reversible.
- **`step`** (`volume up _`) is extracted but not yet wired in v1 — it shows the current value (follow-up; would need a read-before-set).

## Also: brightness `blankProximity` 0 → 3

`brightness 70 _` never matched the brightness keyword (proximity 0 meant only the *adjacent* `brightness _` matched; the inline "70" pushed `_` out of range and it fell through to fluid-blank). Bumped to 3 to match volume, so the inline-value form is reachable.

## Verify

```
volume 30 _     → sets volume to 30 → "30%"
brightness 70 _ → sets brightness to 70 → "70%"
volume _        → still GET (current)
weather london _→ still a lookup (london is the query)
```

- 4 new unit tests: set dispatches `set` then read-back `get`; non-settable degrades; non-numeric degrades; plain get issues no set.
- 2 agentic scenarios (`205`/`206`): set 20→80 and 40→90 — proves the value **actually changes** (a get-only path would still read the old value). Full BlankIntent suite 7/7 (gate + typed-SET + flag-off baseline).
- Runtime suite 1708 green; CC patch-boot + bundle-integrity gates pass.

Off by default (rides `blank-intent-mode`). Built on the merged PR #201.


wkasekende:
**Follow-up commit: Phase-1 loose window (the real proximity fix).**

The first cut bumped brightness `blankProximity` 0→3 as a band-aid. That contradicted the design — BlankIntent is supposed to *replace* per-blank proximity tuning ("the LLM does precision now"). The issue: Phase-1 (`matchKeyword`) still used each blank's tuned `blankProximity`, so the gate only ever saw slots the *old* proximity gate produced — `brightness 70 _` (proximity 0) never reached the classifier.

Now, **when the gate is active, `matchKeyword` widens to one loose window (12 words) for every blank** instead of the per-blank knob. So:
- `brightness 70 _` matches via the loose window (band-aid reverted), → SET → 70%.
- `…brightness…perfect earlier today _` (keyword 8 words back, wouldn't match proximity 0/3) reaches the gate and correctly **CEDEs**.
- Gate **off** → tuned per-blank proximity exactly as before (master behaviour).

This is what makes "proximity tuning no longer needed when the gate is on" actually true. +4 unit tests cover the on/off/unwired/adjacent matrix.

wkasekende:
**Reverted the loose/line-scoped Phase-1 window — it surfaced a real conflict.**

Widening the keyword window in `BlankFill.matchKeyword` alone is unsafe: there are **5 keyword-match sites** that must agree on who owns a `_` — `BlankFill` (script dispatch) plus four core sources (`blank-source` claim, and the `fluid-blank` / `transform-blank` / `config-intent` cede checks), all using per-blank `blankProximity`. Widening only `BlankFill` meant it claimed `brightness 90 _` (loose/line window) while `FluidBlank` did **not** cede (brightness proximity 0) and answered "percent" — a race (it passed by luck earlier; line-scope made FluidBlank win → scenario 206 failed).

Reverted to the consistent state: **per-blank proximity everywhere + brightness `blankProximity` 3** (so `brightness 70 _` matches AND all 5 sites agree). typed-SET is unaffected and 205/206 pass again.

**Proper line-scope is a follow-up:** make all 5 sites line-scoped when the gate is on, via a single shared predicate in `@opencues/core` that both `BlankFill` and the resolver cede/claim checks call — so the window can't drift between them. That's a focused change touching the sensitive resolver cede; worth its own PR, not bolted onto this one.

wkasekende:
**Line-scope, done properly (not just BlankFill).** The previous attempt widened one site and raced; this routes **all 5 keyword-match sites** through a single predicate so they can't disagree.

- New `@opencues/core/keyword-window.ts` — `keywordInWindow(keywordEndIdx, blankIdx, proximity, {lineScoped, lineOf})` + `lineOfWords(text)`. Gate off → `gap ≤ proximity`; gate on → same line as `_`.
- All five call it: `BlankFill.matchKeyword` (runtime claim), `BlankSource` claim, and the `FluidBlank` / `TransformBlank` / `ConfigIntent` cede checks.
- Threaded via a live `blankIntentLineScoped: () => settings.get('blank-intent-mode') === 'on'` getter (resolver → build-sources → each source), so a mode flip takes effect without a source rebuild.
- Brightness `blankProximity` band-aid reverted — line-scope handles `brightness 70 _`.

**Verified:** the race is gone (`brightness 90 _` → SET, no FluidBlank "percent"; scenario 206 stable ×3); resolver cede sources non-regressed (transform/fluid/config-intent/sentence all pass); +12 unit tests; core 911 + runtime 1714 green. PR now bundles: **typed-SET + line-scoped Phase-1**.

PR #214, opened 2026-06-28
feat: unified model dispatch — the sole `_` router (5 heuristics retired)

wkasekende:
**Completes the unified-dispatch arc.** The model classifier is now **always on** (no `dispatch-mode` flag) and **authoritative** over every `_`. One LLM call decides routing and the runtime executes it; the deterministic code stopped deciding and became a safe executor.

## What changed (the simplification)
Five opinionated `_`-routing mechanisms — keyword exact-match, `blankProximity`, `blankReplace: auto`, the source-claim race, BlankIntent — are gone from the live path. They mangled conversational input (`what's the weather like in oslo _` → `what's the Oslo: 21°C…`). Now **one model decision per `_`**: `action` / `lookup` / `transform` / `none`.

- **`dispatch-mode` flag removed** (FEATURES + `OpenCuesState`). The classifier reads the **blanks** LLM bucket; `dispatch-provider:` / `dispatch-model:` overrides remain.
- **Authoritative routing** via `dispatchRoute` on `CueContext`: FluidBlank claims only on `lookup`, TransformBlank only on `transform`, neither on `none`/`action` — and claims **over a colliding keyword**. Fixes the cede-collision where `the capital of france is _` was dropped (both the `countries` keyword blank and fluid ceded → 0 results).
- **Script blanks routed by the model, executed with a clamp floor.** `volume 30 _` / `turn it up a bit _` → classifier picks blank+action+arg; `BlankFill.runScriptAction` clamps `set`/`step` to `[0,100]` and passes the arg as a discrete argv element (never a shell string). Reuses the tested exec/sandbox engine.
- **BlankIntent deleted entirely** — core classifier, `blank-intent-mode`, the 5-site keyword-window lockstep, benches + 14 agentic scenarios — subsumed by this gate. BlankFill cedes every keyword blank to the gate (`gateActive`); keeps only stepValues auto-populate + span preservation.
- **Wired on every host** (cc / oc / gemini / chrome / shell) via `buildSharedRuntime`.

## Floors kept (NOT deleted)
- `determineReplaceMode` (the old `blankReplace: auto` heuristic) survives as fluid-blank's FILL/WIPE **data-loss floor**.
- `blankProximity` survives as the **no-classifier fallback window**.
- The model's replace span is validated/clamped — no buffer destruction, no OOB splice.

Both proximity/auto are **inert on the live path** — opinions removed, safety floors retained ("open judgment → model, data-loss invariant → hardcoded floor").

## Trade-off (accepted)
Every `_` is an LLM call → no offline / no-key operation (degrades to `none`), ~200–1000ms latency softened by the cerebras prefix-cache + a per-buffer-text classify cache.

## Verification
- **Unit suites green:** core 977, runtime 1698.
- **Agentic (OpenCode, cerebras gpt-oss-120b)** — full route matrix: conversational weather → real data + sentence preserved; `capital of france` → Paris (collision fixed); `volume 20 _`/`volume 80 _` → clamped set with state-change proof; `fix typos _ …` → transform; prose mention → no fire.
- CC patch-boot smoke + chrome bundle gate clean.

`@opencues/core` 0.6.2 → 0.6.3, `@opencues/runtime` 0.5.4 → 0.5.5. Design: `docs/architecture/unified-dispatch.md`.


wkasekende:
Abandoning this direction. A centralized LLM classifier in front of every `_` taxes instant deterministic paths, double-classifies sources that already self-classify (causing dropped transforms via the 'buffer changed' skip), and routes deterministic menus/commands worse than master. The two real wins (prose-cede, keyword-collision) are narrow and better solved surgically from master (refine the opt-in BlankIntent gate + the keyword-cede decision). Closing; continuing from master.

PR #246, opened 2026-07-05
sec: close NF1 secret-binding gap on Bun-subprocess LLM path + NF3 dep bumps

gray-raven:
## Summary
- **NF1 (High):** the Bun-subprocess user-blank loader (opencode/shell, PR #148) enforced secret-destination bindings on `ctx.fetch()` but not `ctx.llm()` — a blank could exfiltrate a `secret-hosts`-bound secret via an LLM prompt. `subprocess-loader.ts` now mirrors the in-process `registry.ts` guard.
- **NF2 (Low, defense-in-depth):** `volume`/`brightness` blank scripts now independently reject non-digit input before it reaches an `awk` command-substitution interpolation.
- **NF3:** dependency bumps clearing 9 advisories from a fresh `pnpm audit` — `undici >=7.28.0` and `@babel/core >=7.29.6` pinned via `pnpm-workspace.yaml` overrides; removed the unused `@anthropic-ai/sdk` devDependency from `integrations/chrome` (zero imports repo-wide, pulled in a vulnerable `form-data`).

Full write-up in `docs/INFOSEC_FINDINGS.md` (second-pass section, dated 2026-07-04), including confirmation that every finding from the original 2026-06-07 audit is already fixed.

## Test plan
- [x] `pnpm audit` — 0 vulnerabilities (down from 9)
- [x] `@opencues/runtime` typecheck passes
- [x] Manually exercised `volume-blank.sh set` with injection-shaped and AWK-breakout values — both rejected, normal digits still pass
- [x] `scripts/lint-shell-portability.sh` — all 9 scripts pass
- [x] Confirmed `subprocess-loader.test.ts` failures are pre-existing/environmental (isolated-vm has no prebuilt binary for this dev machine's Node/win32 combo), not a regression from the NF1 fix


wkasekende:
## Follow-up commits pushed to this branch (for your review)

A review pass on this PR found the core NF1 fix **sound** — it correctly resolves the provider hostname and mirrors `registry.ts`'s in-process guard, closing the plaintext-prompt exfil on the opencode/shell path. Pushed four commits on top addressing one real residual + doc/hygiene items. All yours to accept, tweak, or drop.

### 1. NF1 second-pass — type-confusion bypass of the secret scan (`d79f901`)
The Layer-2 literal-value scan is string-coercion based (`` `${system}\n${prompt}` ``), but the user-blank IPC boundary (`subprocess-runner.cjs`) `JSON.parse`s the request with no shape validation. A non-string field —

```js
ctx.llm({ prompt: { x: ctx.secrets.SECRET_B } })
```

— stringifies to `"[object Object]"` in the scan while the real value is JSON-serialized into the wire body downstream, so a bound secret rides to a host outside its binding (the exact multi-secret cross-talk Layer 2 exists to catch). Present on **both** loaders (`registry.ts` + `subprocess-loader.ts`).

**Fix:** both handlers now coerce `prompt`/`system` to strings **once** and forward the coerced values, so the bytes the scan reads are exactly the bytes that reach the wire. Pinned by a new provider-independent regression test (`subprocess-loader.nf1.test.ts`, 3 cases — no isolated-vm / `resolveLLM` dependency). `tsc --noEmit` clean; guard suite green.

### 2. Merge of master + CHANGELOG conflict resolved (`9d28578`)
Branch was CONFLICTING on `CHANGELOG.md` only (both sides added `[Unreleased]` entries). Kept both sets; branch is now MERGEABLE.

### 3. Canonical `security-audit.md` updated (`0273c55`, `dc11ee2`)
CLAUDE.md asks this doc be updated whenever secret-binding code moves. It wasn't — and **row #6 was already 🟢 claiming the `ctx.llm` guard was uniformly in place**, which NF1 shows was false on the subprocess path. Row #6 now records the loader-path gap + closure, the second-pass coercion fix, and stays 🟢 (residual closed; only the out-of-scope provider-logs caveat remains). Open follow-up added for the *optional* shared-helper extraction — NF1 was a drift between the two hand-mirrored builders, and the copied guard re-arms that class; extracting one shared predicate would close it structurally.

### 4. Relocated `docs/INFOSEC_FINDINGS.md` → `docs/architecture/security-findings.md`
It was a standalone dump at the docs root, outside the canonical security home. Its per-finding detail isn't duplicated in the summary table, so it's **moved (not deleted)** next to `security-audit.md` and registered in its Companion deep-dives list. Two in-repo references repointed.

Left unmerged for your approval — would value your read on the coercion fix and whether the shared-helper extraction is worth doing now vs. tracking as follow-up.

gray-raven:
Reviewed all four follow-up commits — the coercion fix is correct (`String()` neutralizes the object-smuggle since the scan and the wire now see identical bytes), the merge/CHANGELOG conflict resolution is clean, and the doc relocation makes sense (`docs/architecture/security-findings.md` is the right home). Manually re-verified the coercion logic against a `toString()`-override variant too — still caught, since the scan reads whatever `String()` produces and that's exactly what gets dispatched.

On the open question: went with the shared-helper extraction now rather than tracking it as follow-up (`69033942`). `buildLlmSecretGuard` in `secret-leak-guard.ts` now owns the hostname-resolution + coerce-then-scan-then-dispatch logic; both `registry.ts` and `subprocess-loader.ts` call it instead of maintaining hand-mirrored copies. Reasoning: this is the second time in one PR that copying this guard across the two loaders caused (or nearly caused) a gap — NF1 itself was fetch-guarded-but-not-llm-guarded drift, and the coercion fix had to be hand-applied twice to close it fully. That's exactly the "control fixed on one path, silently absent on a parallel path" class this repo has hit before (F2/F7, and now NF1 twice) — small enough to fix at the root right away rather than let a third instance happen.

`tsc --noEmit` clean, `secret-leak-guard.test.ts` + `subprocess-loader.nf1.test.ts` pass. Ready for another look whenever convenient — thanks for catching the type-confusion gap, that was a real one.

PR #253, opened 2026-07-05
feat(defaults): ship nav-keymap in the OPENCUES.md template

wkasekende:
## Summary
`nav-keymap` (modifier combo for word nav + cycling: `auto` / `ctrl-alt` / `ctrl-shift`) has existed in the FEATURES registry and the `config _` selector, but the seeded `defaults/OPENCUES.md` never shipped a line for it — so the natural discovery path (reading your settings file) never surfaced it. Found during the website FAQ audit when the maintainer learned the setting existed from his own FAQ.

Adds `nav-keymap: auto` with a preset-options comment block (documenting all three values, the macOS Terminal.app Option-as-Meta note, and chrome's ctrl-alt pin) directly under `cursor-navigate`, matching the template's existing documentation style.

## Behaviour
None changed — `auto` is exactly the registry default that already applied when the line was absent.

## Notes
- `feature-registry-menu.drift.test.ts`'s template check pins MENU_TUNABLES only (FEATURES deliberately excluded), and the shipped value equals the registry `valueOrder[0]` regardless.
- Reaches **new seeds only**: `seed-configs` doesn't retro-add lines to existing users' OPENCUES.md; existing files keep resolving the registry default. If retro-surfacing is wanted, that's a seed-configs self-heal follow-up.

### DIFF
--- a/defaults/OPENCUES.md
+++ b/defaults/OPENCUES.md
@@ -7,6 +7,17 @@ debug-mode: off
 tips-mode: on
 cursor-navigate: inactive
 
+# nav-keymap — modifier combo for word navigation (Left/Right) and
+# alternative cycling (Up/Down).
+#   auto       (default): resolves to ctrl-alt on every host.
+#   ctrl-alt   : Ctrl+Alt+Arrow (Ctrl+Option+Arrow on macOS — in
+#                Terminal.app enable "Use Option as Meta key" so the
+#                combo survives).
+#   ctrl-shift : for terminals that forward ctrl-shift but not
+#                ctrl-alt. Chrome always uses ctrl-alt regardless
+#                (ctrl-shift+arrow extends browser text selection).
+nav-keymap: auto
+
 # max-thinking — how hard reasoning-capable models think before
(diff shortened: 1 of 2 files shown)


wkasekende:
Added verification + tests (6924f81):

**Confirmed working end-to-end:** the template's frontmatter (including the new preset-options comment block) parses through `parseOpenCuesMd` and `nav-keymap: auto` lands in typed state; keymap resolution itself was already covered by `nav-keymap.test.ts`.

**New `opencues-template.drift.test.ts`** pins the shipped template file itself (previously only inline strings were tested):
1. frontmatter yields scalars (anti-vacuity guard)
2. template parses; `nav-keymap` → `'auto'` in typed state, neighbours intact
3. every template scalar is registry-known (plus an explicit allowlist for the hand-edit-only `blank-loading-*` free-text scalars — the test discovered those three ship outside the registry by design)
4. every template FEATURES value is a declared value id

Runs: new file + `config-loader.test.ts` + `nav-keymap.test.ts` = 44/44; core's `feature-registry-menu.drift` (also reads the template) = 11/11.

PR #258, opened 2026-07-05
feat(identity-context): buffer dehydration — safe mode becomes bidirectional (spec 0.5)

wkasekende:
# feat(identity-context): buffer dehydration — safe mode becomes bidirectional (spec 0.5)

## What

`identity-context-mode: safe` previously protected only the catalog direction: IDENTITY.md values never entered prompts (token-only catalog, post-response substitution). But anything personal the user **typed into the buffer** shipped to the provider verbatim — `draft an email _ hi, this is Wilfred from Command Stick` sent both the name and the company.

This PR closes the other half. In safe mode (the default), every LLM-bound copy of buffer text is **dehydrated** — catalog values replaced with their `[TOKEN]`s — before dispatch, and responses are **hydrated** back locally by the existing post-processor (retroactively named). PII no longer reaches the provider whether it came from the catalog or the user's own typing.

Canonical doc: `docs/architecture/hydration-dehydration.md`. Normative contract: `spec/identity-context-spec.md` § Dehydration (spec 0.4 → **0.5**, behaviour-only bump following the 0.3→0.4 precedent).

## Coverage — all 9 outbound channels

| Channel | Hook |
|---|---|
| FluidBlank `INPUT:` + ambient block | dehydrate before `fusedUser`; ambient scrubbed with the same pass |
| TransformBlank `INPUT:` | `extractText` dehydrated before `[CURSOR]` injection (cursor mapped via `mapOffset` — the sentinel can never split a value into leakable fragments) |
| TransformBlank Cerebras `prediction` | ships the dehydrated text (was its own leak channel) |
| SentenceCue `SENTENCE:` | per-span dehydrate; alternatives hydrated back; `alternatives[0]` stays original |
| ConfigIntent classifier + summon | input-only (output is registry-validated); echoed summon phrase hydrated **before** `resolveSummonStart` so offsets stay on original bytes |
| Word-cues | PII words **withheld from dispatch entirely** (`isPiiWord`) — your name gets no LLM synonyms; raw-parser path dehydrates + hydrates |
| AgentRewrite `DOCUMENT:` | windowed doc dehydrated; **hydrate-before-merge** so the three-way merge / DynDefs / cursor math all stay in value space |
| blank-weave `PRIOR TEXT:` | dehydrated; woven output hydrated (`⟦VALUE⟧` weave token untouched — not bracket-shaped) |
| **Defense-in-depth floor** | `dispatchChat` scrubs messages + prediction on every transport; AgentRewrite's HTTP branch (the one dispatchChat bypass) applies the same exported floor. A hit = "a source is missing dehydration" + loud warning. Unregistered (benches/CLI) or `raw` mode → byte-identical pass-through. |

## The mechanism

- New pure/browser-safe core module `packages/opencues-core/src/dehydrate.ts`: one compiled alternation per catalog (longest-value-first, case-insensitive, Unicode-lookaround boundaries with per-side CJK opt-out so CJK names in CJK prose still match), possessive/punctuation variants for free, deterministic single pass (Cerebras prefix cache keeps hitting), `mapOffset` for cursor snapping, `isPiiWord` for the word-cue drop, WeakMap cache keyed on the catalog Map instance (recompiles exactly on IDENTITY.md hot-reload).
- **Visible residual, never silent**: values <3 chars or dictionary/month/day words are skipped and warned once per compile (redacted). A user named June is told, not silently exposed.
- **Round-trip correctness**: `originalBody` passed to hydration is ALWAYS the true pre-dehydration text (pinned by an explicit trap test). New `introducedTokens` option + `report.ambiguous` on `postProcessContext`/`resolveTypedSentinels`: when the user typed a literal token AND its value appears in the same buffer, preserve wins (never inject PII into a deliberate placeholder) and the conflict is reported.
- **No logical landmines**: dehydration only produces outbound copies (buffer never mutated); every hydration site is try/catch'd (worst case: visible token, revertable); AgentRewrite discards the round on hydration failure; throwing-LLM leaves the buffer byte-identical (pinned).

## Bug fixes riding along (both pre-existing, both now security-load-bearing)

1. **`applyOpenCuesScalar` default drift** (`config-loader.ts`): the inline re-parse defaulted `identity-context-mode`/`blank-context-mode` to `off` while the main parse defaults absent → `safe`. Cycling ANY satellite scalar on a config without the explicit key silently downgraded the mode in-memory — post-feature, that would have turned the PII scrub off without a trace. Aligned to two-tier semantics + pinned (unit + scenario).
2. **Resolver identity gate**: `identityContext` was withheld when every `_` was keyword-bound / no `_` present. Word-cue and sentence-cue dispatches now need the catalog to dehydrate, so identity is forwarded whenever mode ≠ off (blankContext keeps the gate — it costs network; identity is an in-memory Map).
3. **Spec staleness**: `spec/identity-context-spec.md` still claimed default-`off`, contradicting `core.md` (default `safe` since PR #161). Fixed in the 0.5 pass.

## Spec bump 0.4 → 0.5

`SPEC_VERSION`, root `SPEC.md`, all spec banners, `spec/CHANGELOG.md` 0.5.0-alpha entry, schema description, new § Dehydration (six normative requirements: coverage, matching floor, visible residual, buffer immutability, round-trip precedence, fail-safe), § Security claims rewritten (Default-safe + Bidirectional). Conformance fixtures for the identity/dehydration surface remain on the tracked TODO in `spec/conformance/README.md` (the suite is still 0.1-alpha flat — pre-existing; the reference impl's executable equivalents are named there).

## Tests

- `dehydrate.test.ts` — 30+ matcher/skip/mapOffset/round-trip pins incl. the originalBody trap + ambiguous-case precedence.
- `sources/dehydration-outbound.test.ts` — per-source negative invariant: **no catalog value substring anywhere in any recorded request body** (messages + prediction) under safe; raw byte-compatible. (Synthetic values — the static prompts contain real-looking example names.)
- `dispatch.test.ts` — floor scrubs+warns / unregistered pass-through / throwing-guard fail-open.
- `agent-rewrite.test.ts` — tokens-out/values-back, user-types-during-call merge invariant through hydration, throwing-LLM byte-identical.
- `identity-dehydration.scenarios.test.ts` — runtime journeys (catalog plumbing from IDENTITY.md on disk, off/raw/missing no-ops, satellite-cycle drift pin, zero token residue in the live buffer).
- Full suites green: core 1024 (node) + 198 (vitest), runtime 1722.

## Follow-ups this PR does NOT cover (deliberate)

- **Live bench re-runs** (need provider keys + same-session baselines per the bench protocol): `fluid-blank-ambient/fused-bench.ts` (target ≥175/176), `transform-blank/prod.ts --provider cerebras` AND `--provider groq` (`--parallel 4`, judge the delta), `tests/benchmarks/user-context/`. The prompts are unchanged, but safe-mode inputs now carry tokens where values were — quality delta must be measured before this merges or immediately after.
- **Agentic-harness pass** (private repo): scenario-103 territory with identity catalogs on — catalog-induced classification effects are invisible to the bare benches.
- **Chrome E2E**: `cd integrations/chrome && npm run build && npm run test:e2e:chrome` — resolver/sources code runs in the content script (silent-degrade class).
- New `tests/benchmarks/dehydration/` quality-delta bench (on vs off) — scaffolding deferred to avoid shipping unexercised live-LLM bench code; the deterministic halves are unit tests.

## Upgrade path

Nothing manual for native hosts: `safe` is already the default, no new config file/scalar; srcHash drift self-heals on the next `opencues run <host>`. Chrome: `npm run build` + `opencues sync chrome` + extension reload (version string moves to 0.2.45). Empty IDENTITY.md = zero behavioural change. Observability: `debug-mode: on` → `grep dehydrated /tmp/opencues.log`.


wkasekende:
## Pre/post-merge TODOs (checklist)

- [ ] **Bench: fluid-blank-ambient** — `npx tsx tests/benchmarks/fluid-blank-ambient/fused-bench.ts`, target ≥175/176. Run a same-session baseline on master first and judge the delta, not the absolute.
- [ ] **Bench: transform-blank prod, cerebras** — `npx tsx tests/benchmarks/transform-blank/prod.ts --provider cerebras --parallel 4` (baseline + delta; cerebras has ~7-case variance at temp=0/seed=42).
- [ ] **Bench: transform-blank prod, groq** — same with `--provider groq --parallel 4` (watch for rate-limit "(bailed)" phantom regressions).
- [ ] **Bench: user-context suite** — `tests/benchmarks/user-context/` (the sentinels design-evidence corpus; safe-mode inputs now carry tokens where typed values were).
- [ ] **Chrome E2E** — `cd integrations/chrome && npm run build && npm run test:e2e:chrome` (resolver/sources run in the content script — silent-degrade class).
- [ ] **Agentic harness pass** — scenario-103 territory with identity catalogs ON; add an absence-of-regression assertion (debug-log `notMatches <raw catalog value>`). Catalog-induced effects are invisible to the bare benches.
- [ ] **Live probe on the CC fork** — populate IDENTITY.md, `debug-mode: on`, type your real name + `fix typos _ …`; confirm `grep dehydrated /tmp/opencues.log` fires and the buffer round-trips with the real value intact.
- [ ] **Chrome mirror sync** — after merge: `npm run build` + `opencues sync chrome` + reload, confirm 0.2.45 in `chrome://extensions`.

## Follow-up issues to file (not blockers)

- [ ] **Pack `ctx.llm` floor gap** — packs' capability-gated LLM access does its own HTTP outside `dispatchChat`, so the outbound PII floor doesn't cover it; captured blank args can carry user-typed text. Either dehydrate captured args before pack context, or apply `applyOutboundDehydrationFloor` inside `ctx.llm`. Documented in `docs/architecture/hydration-dehydration.md` § Known residuals.
- [ ] **`tests/benchmarks/dehydration/` quality-delta bench** — on-vs-off end-to-end quality measurement (deliberately not scaffolded in this PR to avoid shipping unexercised live-LLM bench code; deterministic halves are unit tests).
- [ ] **Conformance fixtures** — `dehydration/` round-trip + matching-floor fixtures added to the tracked TODO in `spec/conformance/README.md` (suite is still 0.1-alpha flat; pre-existing).
- [ ] **security-audit row #19 staleness** — found during the docs pass: the "10%-length" TransformBlank guard the row describes appears retired (`detectPartialTransform` imported but no call site found; live guard is AgentRewrite's `validateLLMRewrite`). Unrelated to this PR — needs its own cleanup.

PR #272, opened 2026-07-07
fix(cc): pin tweakcc to v4.3.0 + native-binary startup smoke in validateFork

wkasekende:
## The bug

`opencues install claude-code` **broke `claude-cues`** — the reinstalled native binary crashed on startup and OpenCues never booted. Root cause:

`setup.sh` cloned `github.com/Piebald-AI/tweakcc` at **HEAD**, which pulled the day-old **4.3.1** (released 2026-07-06). tweakcc 4.3.1 regressed the native-binary `.bun`-section repack: the repacked `claude.exe` bloats **247MB → 411MB** and dies at startup with:

```
TypeError: Expected CommonJS module to have a function wrapper.
Bun v1.3.14 (Linux x64 baseline)
```

Bun can't load the corrupted embedded module graph → no `[cc]` log lines → the agentic harness bridge never arms. The extract + JS patch were **fine** (16.58MB → 16.62MB, just the +40KB OpenCues bootstrap); only tweakcc's *repack* broke.

**And it shipped silently:** `validateFork` reported the dead fork "installed + validated" because its boot-smoke only `require()`s the dist modules under **Node** — it never runs the actual **Bun binary**. So the corrupt repack passed validation.

## Bisection (via git tags)

| tweakcc | binary | boots? |
|---|---|---|
| 4.3.1 (HEAD) | 411MB | ❌ crashes |
| **v4.3.0** | 411MB | ✅ **boots** |

The 164MB bloat is *normal* for tweakcc's native repack (v4.3.0 has it too and boots fine) — the regression is a module-graph corruption in the 14 commits after the `v4.3.0` tag. **v4.3.0 is the newest good version.**

## Fix

1. **`compat.json`** — add `tweakcc-pin: "v4.3.0"` (single source, next to `current-pin`) with a note explaining the 4.3.1 regression.
2. **`setup.sh`** — clone then `git checkout` the exact pinned tag (read from `compat.json`, `TWEAKCC_REF=<tag>` env override for bisecting a future release), **not HEAD**. Checkout runs before `npm install` so deps match the pinned tree; aborts loudly if the tag doesn't exist.
3. **`install.cjs:validateFork`** — native-binary **startup smoke**: after patching, `spawnSync(claude.exe, ['--version'])` must exit 0 and print `Claude Code`, else the fork is refused with a pointer at `compat.json:tweakcc-pin`. (cli.js shape is plain Node JS, so this failure mode is native-only.) This is the guard that would have caught the regression.

`@opencues/claude-code` 0.2.7 → 0.2.8; root CLAUDE.md § Common drift-bug patterns documents the class.

## Verification

- `opencues install claude-code --rebuild` → `▸ Cloning tweakcc (pinned v4.3.0)` → `● claude-code (native) installed + validated.` → `claude.exe --version` = `2.1.170 (Claude Code)`.
- The agentic harness bridge **arms** on the rebuilt binary (core 0.18.0).
- The new smoke is proven to discriminate: good v4.3.0 binary → passes; a crashing binary → fails.
- `bash -n setup.sh`, `node -c install.cjs`, legacy-names + shell-portability lints all clean.

## Note

I hit this while trying to run the CC-scoped agentic cycling scenarios after reinstalling CC to core 0.18.0 (the Linked Words removal). The reinstall exposed the pre-existing unpinned-clone landmine; it was **not** caused by the 0.18.0 core (opencode ran the full suite fine on it). Your `claude-cues` is restored and working with OpenCues at 0.18.0 on v4.3.0.


wkasekende:
Superseded by #277, which landed the same two fixes (tweakcc pin in compat.json + startup smoke on the patched artifact in validateFork) plus the fatal cli.js node --check, the missing-marker-is-stale health-gate closure, and a CI gate (check-tweakcc-pin.sh) keeping all of it in place. The 4.3.1 repack regression this PR bisected is also covered by #277's runtime smoke — and per the CC 2.1.206 validation in #278, the module-graph corruption is fixed on upstream tweakcc HEAD as of 1545ff8 (Jul 10): the repacked binary boots and passed the agentic suite.

PR #292, opened 2026-07-13
feat(undo): language-invariant `undo _` / `redo _` — revert anything OpenCues did

wkasekende:
## What

Typing `undo _` reverts the last change the runtime made — a blank fill, a fluid/transform substitution, an agent-rewrite round, a cycling step, a settings write, a volume/brightness set, an IDENTITY.md/NOTES.md write. `redo _` re-applies; `undo 3 _` (or "undo the last two changes", `3回元に戻して _`) reverts deeper. ON by default (`undo-mode: off` disables).

Docs: `docs/features/undo.md` (user-facing) + `docs/architecture/undo.md` (canonical reference, incl. § "What's NOT in scope").

## Design (agreed scope: Tiers 1–3 implemented, Tier 4 documented)

**Routing — a fourth verdict kind, not a keyword blank.** `ActionVerdict {action, count}` joins SETTING/PROVIDER/NONE on the config-intent classifier. The system prompt stays byte-stable (cerebras prefix caching); `fluid-config-mode` and the new `undo-mode` scalar gate their verdict kinds post-classification. Action verdicts carry **no emit-time side effect** — core classifies, the runtime applies. The likely-intent gate gains a multilingual undo/redo alias list, and `hasLikelyIntent` now substring-matches non-ASCII tokens (JS `\b` is `\w`-based and never matches CJK/Cyrillic/Thai — a silent language filter).

**UndoJournal (session-scoped transaction log).** Typed entries: buffer diffs via a new `diffSplice` helper (no span-shift plumbing — relocation by unique match at apply time), scalar writes with prior values, OS sets captured get-before-set, sentinel/note writes as **validated inverse blank-ops** (exposed through `createBlankInvoke`'s `ProcessResult.writeInverse` — zero integration-bootstrap changes; the inverse replays through `validateSentinelWrite`/`validateNoteWrite` by construction, audit row #24 intact). Cycling bursts coalesce (volume ×6 → one undo). Buffer resets bump an **epoch** instead of wiping — settings/OS changes from a previous message stay undoable; stale text entries skip with a report.

**UndoApplier — exact-match-or-refuse.** Unique-match relocation for text; verify-then-set for scalars (via the extracted `applyScalarAndPersist` — reload suppression arms) and OS values (get first; drift → skip). Partial failure is reported (statusline `undo` block + `undo.applied` event), never masked; consumed transactions pop regardless so the stack never wedges; one `pushText` per apply (one host history entry). Nothing applied → inline `[OpenCues: nothing to undo]` with a clearOnEdit span. Tier-4 (user-pack fetch/exec effects) journaled as `external` — buffer reverts, effect declared irreversible.

**Refactor:** cycling's `invokeOrSpawn` + the resolver's `applyOpencuesScalar` wrapper extracted to shared utils so the applier can't drift from the live pairs (the copied-guard bug class).

Wired on all six adapter bands (incl. CC's inline construction). Runtime-only feature — no SPEC_VERSION bump, no reserved blank name. `@opencues/core` 0.18.1→0.19.0, `@opencues/runtime` 0.15.1→0.16.0, chrome 0.2.65→0.2.66 (lockstep — the runtime lands in its bundle).

## Verification

- **Unit/scenario**: 20 journeys in `undo.scenarios.test.ts` (relocation, refusal honesty, coalesced burst revert, epoch skip + sibling revert, drift skips, race bail, redo semantics, reentrancy) + `undo-journal.test.ts` (20) + classifier parse/validate/gating tests + build-sources flag-combination tests. Full core (1098 node + 295 vitest) and runtime (2012) suites green.
- **Bench** (`tests/benchmarks/fluid-config/prod.ts`, NEW — drives the production SYSTEM_PROMPT; `fused.ts` only certifies its own settings-only prompt): same-session baseline vs new, cerebras `--parallel 4`. Settings: precision 100%/100%, recall 84.8%/84.8% (in-prompt), 60%→65% (holdout) — **no regression** from the added intent. Undo suite: **100% precision / 93.3% recall** (multilingual incl. ja/zh/es/de/fr/ru + number-word counts). Log: EXPERIMENTS.md § v2.2.
- **Chrome E2E**: 16/16 passed against the rebuilt bundle (trust-gate, sensitive-field, site-filter all live).
- **pre-pr.sh**: 11/13. The two failures are not from this diff: `test-hermeticity` fails on two pre-existing WSL-environment-dependent CLI tests (`opencues sync --wsl` / `opencues which` — zero diff lines in `packages/opencues-cli`), and `doctor` reports this machine's stale installs vs the just-bumped source versions (expected until post-merge `opencues install`).

## Upgrade path

`opencues install <host>` / next `opencues run <host>` (srcHash self-heal). No new config file; `undo-mode` defaults ON even without an OPENCUES.md line (deliberate `!== 'off'` polarity). Chrome: normal bundle rebuild + `opencues sync chrome`. Manual smoke: `weather london _` → fill → `undo _`; `voice mode off _` → `undo _` (check OPENCUES.md); volume step ×3 → `undo _` (volume restored).

Follow-up (not this PR): agentic-harness scenarios in the private repo.

https://claude.ai/code/session_01Lun2cr5zEUEv5GaZDMpLrX


wkasekende:
Hardening pass after review ('add all of them'): tap-assertion journeys + agentic scenarios + live fork run. **Three real bugs found and fixed before merge**, each invisible to the original test layer:

1. **Coalesce frame corruption** (`ecc6eee7`) — cycling taps recorded diff-trimmed slices, so successive cycles lived in different frames and the coalesce merge composed garbage ('the lattorney filed' on undo). Fixed: exact-range slices at the taps + a `framesCompose` guard in the journal. Caught by the new `undo-taps.scenarios.test.ts` (10 journeys driving the real cycling/BlankFill/registry paths, incl. note/sentinel file-write round-trips through their validators).
2. **Post-confirmation misclassification** (`50804416`) — `debug-mode on undo _` (the buffer state after ANY settings flip — the dominant live undo context) classified as SETTING and re-applied the change. Fixed with an additive prompt rule + few-shot; 5 new bench cases; cerebras re-run: undo recall 90%, settings precision 100% intact (EXPERIMENTS.md § v2.3). Caught by agentic scenario 109.
3. **Summon-span anchor swallow** (`f3a3f51b`) — the summon resolver claimed the whole buffer as the command, splicing away the applier's relocation anchor (buffer emptied on live CC). Fixed with deterministic span tightening (never swallow a pending transaction's anchor) + two whole-buffer-span regression tests.

**Live verification**: 3/3 agentic scenarios pass on the patched CC native binary against cerebras (opencues-agentic#16). Full suites: runtime 2027, core 1098+295, all green.

PR #302, opened 2026-07-14
fix(windows): verify-before-restore + verify-before-write in the shim's paste/animation paths

wkasekende:
## The two live incidents (2026-07-14, same session)

**1. Your clipboard pasted into Discord.** `PasteReplace` = save clipboard → set substitution → Ctrl+V → `Thread.Sleep(300)` → restore. Electron consumes the clipboard **asynchronously and unboundedly** — the trace shows Discord's write bracket open 1094ms after the paste keystroke — so the restore won the race and Ctrl+V delivered the user's *old* clipboard (`Izzy.roser@gmail.com`) into the field instead of the substitution. This is a **clipboard leak into whatever app is focused**: swap the copied email for a password or 2FA code and it lands in a Discord input box.

**2. Typed characters eaten.** `congratulations letter _` arrived at the daemon as `gratulations letter _`. `TryTypeMicroEdit` (animation frames) and `PasteReplace`'s backspace path both diff against the shim's *model* of the field (`_lastSentText` / daemon `oldText`); Phase 1 has no keyboard hook, so user keystrokes are invisible between event reads — a stale model turns the backspace burst into user-content deletion.

## The fix — one primitive, three sites

New `TryReadCurrentField` (on-demand, mode-aware: `TryReadFocusedElectron` for MSAA, `ReadValue(_hookedEl)` for UIA):

- **Verify-before-restore**: after Ctrl+V, poll the field until it reflects the pasted text (EolNorm-folded — apps re-dress newlines, per the branch's rendering table), *then* restore the saved clipboard. On timeout (`OPENCUES_CLIPBOARD_RESTORE_MAX_MS`, default 3000ms ≈ 3× worst observed) or unreadable field: **fail safe — skip the restore and warn.** Losing the old clipboard contents is an annoyance; pasting them into the foreground app is a leak.
- **Micro-frames drop on divergence**: read-before-write against `_lastSentText`; mismatch = user typed (or app edited) → swallow the frame. Animation is cosmetic and the final substitution write is an absolute anchor, so a dropped frame costs nothing. Known trade: on a very slow editor, a frame whose predecessor hasn't rendered yet reads as divergence and drops — the animation stalls a frame, never a wrong write.
- **PasteReplace rebases on a fresh read**: `oldText` superseded by the live field before computing the backspace count, killing the eaten-characters class on the paste path too.

All C#5-compatible (Add-Type compile at launch; pre-declared out-vars, concatenation, matches file style).

## Guarding the class

`tests/clipboard-invariants.mjs` — source-invariant guard in the same pattern as `newline-invariants.mjs` (the shim can't be unit-tested off-Windows; the invariants whose loss is silent get pinned instead): consumption-gated restore, no timer-raced restore shape, loud fail-safe, fresh-read rebase, micro-frame divergence drop, EolNorm-folded comparisons. **6/6 green**, wired into `pre-pr.sh` next to the newline guard. Existing newline invariants still ALL PASS.

`@opencues/windows` 0.1.0 → 0.1.1 · windows CLAUDE.md gains a "verify-before-write + verify-before-restore" section documenting the incident pair.

Manual verification checklist (Windows-side, next launch recompiles the shim): (1) copy a known string → trigger a big substitution in Discord → confirm the substitution lands AND clipboard still holds the string (or the warn fires and clipboard holds the substitution — never the reverse paste); (2) type continuously while a slow blank resolves → no leading characters lost; (3) `grep 'micro-frame skipped\|clipboard NOT restored\|rebasing diff' /tmp/opencues.log` to see the guards working.

### DIFF
--- a/integrations/windows/CLAUDE.md
+++ b/integrations/windows/CLAUDE.md
@@ -173,6 +173,39 @@ declines and falls through to the whole-value path — where
 big, non-tail write) always takes a whole-value path — typing is only
 for the small frames.
 
+## Verify-before-write + verify-before-restore (2026-07-14 incident pair)
+
+Two silent races shipped in the write paths and both fired live on the
+same day; the fixes share one primitive, `TryReadCurrentField` (an
+on-demand, mode-aware read of the attached field), and are pinned by
+`tests/clipboard-invariants.mjs` (in pre-pr):
+
+1. **Clipboard restore raced the app's async paste read.** `PasteReplace`
+   saved the user's clipboard, set the substitution, sent Ctrl+V, slept
+   a fixed 300ms, restored. Electron consumes the clipboard
+   asynchronously — Discord under load was observed reading >1.1s after
+   Ctrl+V — so when the restore won, the paste delivered the USER'S OLD
+   CLIPBOARD into the focused app (live: a copied email address landed
+   in a Discord input instead of the substitution). That is a clipboard
+   LEAK into whatever app is focused. Now: poll the field until it
+   reflects the pasted text (EolNorm-folded), THEN restore; on timeout
+   (`OPENCUES_CLIPBOARD_RESTORE_MAX_MS`, default 3000) FAIL SAFE — skip
+   the restore and warn. Losing old clipboard contents is an annoyance;
+   pasting them into the foreground app is a leak.
+
+2. **Write diffs computed against a stale field model.** Both
+   `TryTypeMicroEdit` (animation frames) and `PasteReplace`'s backspace
+   path diffed against the shim's belief (`_lastSentText` / the
+   daemon's `oldText`). Phase 1 has no keyboard hook, so USER keystrokes
+   are invisible between reads — a stale model turns the backspace burst
+   into user-content deletion (live: "congratulations" typed mid-
+   animation lost its leading "con"). Now: read the field immediately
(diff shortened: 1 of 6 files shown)


wkasekende:
Live validation results (user test, 18:35):

- **Leak: dead.** Big Discord substitution landed correctly; the fail-safe fired (`clipboard NOT restored … after 3000ms` warn) instead of racing the restore — nothing foreign pasted. Micro-frames all applied cleanly; the post-paste `_` re-fire suppression worked.
- **But the fail-safe fired on a paste that plainly landed** (bracket reconciled 47ms after the warn): Discord/Slate re-dress pasted text in accValue readbacks, so EolNorm equality never matches there — the user's clipboard restore would be sacrificed on *every* big Discord substitution. Safe, needlessly lossy.

Follow-up commit `7194b33e`: **AlnumFold fallback** on the consumption match (readback must contain a 24-char lowercase-alphanumeric prefix window of the pasted text; gated at ≥12 fold chars so short generic pastes can't false-match). Exact match stays the fast path. Invariants now 7/7 (`clipboard-invariants.mjs` pins the fallback + gate); `clipboard-restore.manual.md` added with the Windows-side checklist (leak repro, eaten-keystrokes repro, and an anti-lossiness rate check — a 100% warn rate in any app means its readback dress needs cataloguing).

PR #304, opened 2026-07-15
fix(blank-fill): defer the registry-miss [err] fill past the host's text-state catch-up

wkasekende:
## Caught live during slot-2 validation of #300

On real CC the registry-miss guard half-worked: the once-per-blank **warn fired**, but the `[err]` fill **never appeared** — `loading animation ▖,▘,▝,▗ _` stayed untouched. Trace: slot formed → invoke → warn → silence, no `substituting` line.

Root cause: the miss branch is the **only** fill that ran *synchronously inside the text-change dispatch*. On CC, `adapter.getText()` is one keystroke stale at that instant (the host's buffer state catches up after the handler returns), so `words[slot.index]` missed and `applyAsyncFill`'s staleness guard silently dropped the fill. Every other fill path is naturally deferred past the state update by script/LLM latency — and the MockAdapter commits text **before** dispatch, so the #300 unit journeys couldn't see the class (they passed honestly against a host model that doesn't exist on CC).

## Fix

Defer the `[err]` fill one tick, with one guarded retry at 50ms: `tryErrFill` probes for the slot's `_` (or its animation frame char) before filling; if the buffer legitimately moved on instead, the retry is dropped by the same staleness guard that protects late script callbacks. No behaviour change for any other fill path.

## The harness gap, closed structurally

`MockAdapter` gains **`staleTextDuringDispatch`** — dispatch handlers with the new text in the *event* while `getText()` still returns the pre-change buffer, committing after dispatch returns (the live-CC shape). A new journey pins the `[err]` landing on such a host, and it's **mutation-checked**: fails against the synchronous code, passes with the defer. Any future fill path that runs synchronously in-dispatch is now testable against this host class.

Full runtime suite green (2 unrelated load-flakes — brightness script smoke + seed-configs SEED — pass in isolation, matching their history). `@opencues/runtime` 0.16.1 → 0.16.2 · CHANGELOG updated.

Live verification on a real CC PTY follows in a comment before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed — the registry-miss `[err]` fill was dropped on hosts whose text state lags the dispatch (`@opencues/runtime` 0.16.1 → 0.16.2)
+
+Live-CC follow-up to the previous entry, caught during slot-2 validation: the warn fired but the `[err]` never appeared. The miss branch was the ONLY fill running synchronously inside the text-change dispatch, and CC's `adapter.getText()` is one keystroke stale at that instant — `words[slot.index]` missed and `applyAsyncFill`'s staleness guard silently dropped the fill. Every other fill path is naturally deferred past the state update by script/LLM latency, and the MockAdapter commits text BEFORE dispatch, which hid the class from the unit journeys.
+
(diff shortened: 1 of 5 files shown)


wkasekende:
Live verification on a real CC PTY (the host that broke it), post-fix build: injected `loading animation ▖,▘,▝,▗ _` → buffer = `loading animation ▖,▘,▝,▗ [err] loading-animation: not available on this host — stale bundle or missing prerequisite. Try \`opencues install claude-code\``. Command preserved, only the `_` replaced. The pre-fix build on the same host: warn only, buffer untouched.

PR #332, opened 2026-07-24
security(core): ground contradiction-cue geocoder inputs + cap calendar feed size

gray-raven:
## Summary

Static security review of the July 2026 external-HTTP feature cluster (calendar `.ics` ingest + Tier-0…5c contradiction cues). Both surfaces are opt-in (`contradiction-cues-mode` off by default; calendar needs a configured feed), so neither was a live default-on exposure. Two hardenings, both scoped to `@opencues/core`.

## Findings

**1. Journey cue sent ungrounded LLM output to an external geocoder (Medium — fixed).**
`contradiction/checks.ts:verifyJourneyClaim` grounding-checked only `claim.quote`, then geocoded `claim.origin`/`claim.destination` — LLM-emitted strings never verified against the sentence — via `photon.komoot.io`. This is the single place the load-bearing "no LLM-output → side-effect channel" invariant (`security-audit.md`) was touched, and it violated the layer's own stated grounding rule. A prompt-injection could route arbitrary text to the third-party geocoder; a hallucinated place could fire a false cue.
Fix: both endpoints must now be a short (`≤ GEOCODE_NAME_MAX = 80`) literal substring of the sentence (`isGeocodableName`) before either is geocoded — only strings already present in the buffer the LLM was given ever leave the machine. The egress host was and remains hardcoded (never LLM-chosen); chrome additionally gates it through the SW's exact-origin `FETCH_ALLOWED_ORIGINS` allow-list, so this was never arbitrary-host SSRF.

**2. Calendar feed sync had no response-size bound (Low — fixed, defense-in-depth).**
`syncCalendarFeeds` now drops any single feed body over `CALENDAR_FEED_MAX_CHARS` (5 MB) before parsing, bounding memory/parse work on a pathological or MITM'd response (feed URLs are user-configured — user-against-self).

## Reviewed and found solid (no change)
`ics.ts` (bounded, never throws, invalid TZID caught); weather/TfL/bank-holiday providers (fixed URLs or `encodeURIComponent`, defensive parsing, TTL-gated); `safeEvalArithmetic` (shunting-yard, no `eval`); the refresh scheduler (unref'd, in-flight guarded); chrome's SW fetch relay (`isInternalSender` + exact-origin allow-list, F6 intact).

## Dependency audit
`pnpm audit`'s 3 high advisories (`immutable` via `@types/draft-js`, `brace-expansion` via the shell build-time babel plugin) are all transitive dev/build/type-only deps that never execute in shipped runtime (matches the existing DA3 treatment). A clean `pnpm.overrides` fix isn't achievable without broad lockfile churn / breaking `@types/draft-js`'s pinned 3.x API — recommend applying on Linux/CI where `pnpm audit --fix` can be validated. Not runtime-reachable, low urgency.

## Tests
`@opencues/core` full suite green: **1254 `node:test` + 298 vitest, 0 failures**, including 3 new tests (2 `SECURITY:` journey grounding, 1 calendar size cap). Changes are pure-TS and isolated to core; no runtime/cli source touched.

- `security-audit.md` rows #28/#29 added; CHANGELOG Security entry
- `@opencues/core` 0.32.0 → 0.33.0


wkasekende (approved):
Reviewed with a fresh local build — full core suite green (1276 node:test + 298 vitest). The geocoder grounding is the right fix in the right place, and I independently verified gate coverage: the only other `geocodePlace` call site (`homeBias`) takes its input from the system-timezone→city table, not LLM output, so it's grounded by construction. The spy-fetch SECURITY tests follow the house negative-invariant style.

Two notes for the record, neither blocking:
1. The 5 MB calendar cap is post-download — `await res.text()` buffers the full body before the length check, so it bounds parse work but not download memory. Given feeds are user-configured (user-against-self), acceptable; a streaming read would close it fully if we ever care.
2. Version coordination: this PR's core 0.33.0 collides with #334's claimed 0.33.0 — #334 will renumber to 0.34.0 in its next master merge.

PR #345, opened 2026-07-26
test(harness): ambient: inject command for field-declared WIPE + on-field coverage

wkasekende:
## What

Adds an `ambient:` command to the agentic event bridge (`packages/opencues-runtime/src/event-bridge.ts`) that injects a synthetic `AmbientContext` the resolver reads via `adapter.getAmbientContext`:

- `ambient:{"singleLine":true,"app":"chrome"}` — declare a single-line search box / omnibox
- `ambient:null` — clear (identical to a native prose host)

## Why

Two recently-merged features are **field-declared** — their behaviour is driven by the host's `AmbientContext`:

- **#342 field-declared WIPE** — fluid-blank replaces the whole field only when the host declares `singleLine`/`disposable`
- **#343 `on-field:` scoping** — a prose cue cedes in a `single-line` field

Production hosts supply ambient from the real focused field, but the headless harness has no field, so **neither feature had any agentic coverage**. This bridge command closes that gap. The override installs once and falls through to the adapter's own `getAmbientContext` when nothing is injected, so a host that genuinely reports ambient is unaffected until injected. Test-only; requires `ambient-context-mode: on`.

## Verification

Two new scenarios in the private harness repo, both **2/2 stable across re-runs** (anthropic/claude-haiku):

- `119-fluid-blank-field-declared-wipe` — asserts the runtime's own `mode` decision on `fluid-blank.completed` (WIPE under `singleLine`, FILL with no ambient), never LLM text. The FILL control proves WIPE is declaration-driven, not always-on.
- `120-on-field-cede-single-line` — control-first: `more-formal` lands a passive `sentence-cue:more-formal` def in a multi-line field, then cedes to an empty def list under the SAME sleep budget in a single-line field — so absence means the gate ceded, not that the LLM was slow. The `supports()` gate is synchronous (no LLM call on cede), so the assertion is deterministic.

`@opencues/runtime` 0.28.0 → 0.28.1 (additive test helper). Version-bump gate clean.

https://claude.ai/code/session_01XN1qzJAvL12sbkwnLELNkQ


wkasekende:
Folded into #346 (retargeted to master so CI runs — GitHub only runs the pull_request workflow on base=master). Both commits (the `ambient:` harness inject + the field-shape ungate) live in #346 now.

PR #347, opened 2026-07-26
fix(chrome): stacked _ on non-cycling <input> silently ignored

wkasekende:
## Symptom

On a plain **non-cycling** chrome field (Luma's form fields, any normal `<input>`/`<textarea>`), the **first** `_` fills correctly but every **subsequent** `_` is silently dropped — no fill, no log line.

## Root cause

After a fill, the runtime's own view of the buffer goes stale:

- Normal-input writes go through `writeNormalInputValue`, which dispatches a **synthetic** `input` event with `isTrusted=false`. content.ts drops those for security, so the fill never round-trips back to the runtime.
- `runtimeRender()` **early-returns for normal inputs** (they can't host CSS Custom Highlights), so the render path that *also* refreshes the buffer baseline never fires either.

With both refresh paths skipped, the chrome adapter's `lastSeenText` **freezes at the pre-fill value** (the first `_`). The resolver's explicit-`_` gate reads that as `previousText`, so the next `_` computes:
- `blankJustTyped = text.endsWith('_') && !prev.endsWith('_')` → **false** (prev still ends with `_`)
- count-diff fallback `newU > prevU` → **false**

…and the trigger is discarded *before* the gate's own `BLOCKED` log — which is why it's completely silent (confirmed: zero `explicit-_ gate BLOCKED` lines in the user's log despite the `_` being ignored).

This is the same **"stacked-blank after substitute silently no-ops"** class the `e.previousText` gate was built to fix — but that gate relies on the adapter keeping `lastSeenText` current, which the normal-input write path didn't.

## Fix

Wrap `setText` in the chrome adapter band (`adapters/chrome/v1/boot.ts`) so every runtime write updates `lastSeenText`/`lastSeenCursor` — a pure variable write, no event emit, **no re-entrancy**, covering every host write path (cycling or not). Managed-editor writes already round-trip a trusted `input` event, so this only changes the previously-broken normal-input case.

## Verification

- Diagnosed from the user's `/tmp/opencues.log` (three-host interleaved): every fill was `textLen=1`, **zero** `gate BLOCKED` lines, `lastSeenText` never refreshed after a normal-input fill. Each code path in the chain confirmed by reading source.
- Built runtime + chrome, synced to the Windows extension path; user verifying live on Luma (repeated `_` in the same field).

**Testing note:** this is a chrome content-script + adapter integration bug — the agentic harness (opencode/CC) can't reproduce it (those adapters round-trip writes correctly), and `boot()`'s wiring doesn't expose a text-subscriber for a clean unit test. The correct coverage level is the chrome e2e (`integrations/chrome/tests/e2e/`); an e2e case for repeated normal-input fills is an owed follow-up.

`@opencues/runtime` 0.28.2 → 0.28.3, `@opencues/chrome` 0.2.90 → 0.2.91 (manifest + package.json lockstep — bundle bytes changed).

https://claude.ai/code/session_01XN1qzJAvL12sbkwnLELNkQ


wkasekende:
Superseded by the clean fix — the real root cause was the loading-animation '_' frame poisoning the sourceReclassifier (not previousText staleness). See the new PR; this branch's write-wrapper/focus-sync were chasing the wrong layer.

PR #395, opened 2026-08-17
feat(dsh): OpenCues as a first-class DeepSeek Harness plugin

wkasekende:
> **Stacked on #393** (the `harness` provider it routes through). Base is `feat/harness-llm-provider`, so review that first; GitHub will retarget this to `master` once #393 merges. Companion docs PR: #394.

```
dsh plugin --profile web add @opencues/dsh
```

That is the whole install. This is the **first integration that is a real plugin of its host** rather than a patch or a fork — no version pin, no binary surgery, no separate host process, no `opencues install` step required.

[DeepSeek Harness](https://github.com/deepseek-ai/deepseek-harness)'s composer turns out to already be an overlay editor (a transparent `<textarea>` over a backdrop of real DOM text nodes), so `adapters/chrome/v1` drives it unchanged and the CSS Custom Highlight API paints per-word ranges with no DOM mutation of ours.

## Distribution, and why it's npm

Researched against their [publish guide](https://github.com/deepseek-ai/deepseek-harness/blob/master/docs/user/develop/basic/publish.md). `dsh plugin add` forwards to pnpm, so npm / `github:` / tarball / local path all work. **npm with prebuilt artifacts is the only one that asks nothing of the user**: a `github:` install fetches sources, needs a `prepare` script, and requires the user to add the package to `allowBuilds` in their profile's `pnpm-workspace.yaml` — which their own docs describe as *"permission to execute the package's code on your machine at install time, outside any sandbox"*. We ship artifacts rather than ask for that.

So `@opencues/dsh` is **publishable**, and structurally has to be — an unpublished package means no install path at all. It carries **no `dependencies`**: `build.mjs` inlines `@opencues/{core,runtime}` into `client.js` at publish time, which is exactly what lets those two stay `private`.

Discoverability is the `dsh-plugin` GitHub topic (their only official channel alongside Discussions); the community aggregators index from that topic, not from package names, so the `@opencues` scope costs nothing there.

**Not published by this PR.** The tarball is verified but publishing is yours to run.

## Browser-ness became a property instead of a name

Adding `dsh` gave us a second browser host, which exposed that several behaviours were spelled `hostName === 'chrome'` while chrome was the only one: the **ctrl-alt keymap** (because ctrl-shift+arrow is the *browser's* own extend-selection-by-word) and **`dim-mix`** (because the dim colour mixes toward a *page* background). Both follow from being in a browser, so `BROWSER_HOSTS` + `isBrowserHost()` carry them now and a new browser host inherits by construction. Left as a name comparison, dsh would silently have got a keymap the page steals from it — navigation appearing to do nothing, no error anywhere. `statusbar-position` stays chrome-only on purpose: that one is about a surface chrome draws.

A test asserts every host is **exactly one** of native/browser, so a future host added to neither list fails rather than quietly inheriting terminal semantics. No `SPEC_VERSION` bump — `spec/core.md` declares `HOSTS` non-normative and permits runtime-defined names, which `windows` already relies on.

## CLI

`opencues install dsh` / `uninstall dsh` / `run dsh` (+ `deepseek` aliases), a `dsh` row in `doctor`, shell completion.

The installer is a **thin wrapper that hands off to `dsh plugin add`** rather than writing into `$DSH_HOME/profiles/` itself — that directory is dsh's to manage, and reaching into it would be the kind of second install path that later disagrees with the first. Its `--help` says plainly that most users want dsh's own command. `opencues run dsh` earns its keep on one step before the exec: `ensureFreshBundle` rebuilds a stale bundle, which matters more here than on terminal hosts because the bundle inlines core+runtime and is served per page load, so staleness is otherwise completely silent.

## Verified against the real published shape

`npm pack`, then `dsh plugin add ./opencues-dsh-0.1.0.tgz` into an **empty `$HOME` with every OpenCues provider key stripped from the environment** (keeping only `DEEPSEEK_API_KEY`, which a dsh user has by definition):

- installs in **1.5s with no build prompt** and no `allowBuilds` grant
- 394 KB tarball, 6 files, zero dependencies
- loads **29 baked defaults**, builds **7 sources**
- `the capital of iceland is _` → `Reykjavik`, `i has three cats fix typos _` → `I have three cats`, on DeepSeek's model with no OpenCues key
- a settings write creates `~/.cues/OPENCUES.md` from the shipped default, and the blanks still resolve after a reload

## Two incidental test fixes

Both assertions pinned the host list as one comma-joined substring, so they failed on a host being **added** — not a regression, and the failure named nothing useful. Now per host, plus a new test pinning `install.cjs`'s pre-build fallback host list against core's `HOSTS` (drift there means `install <new-host>` works normally and then reports "unknown host" on any machine that hasn't built core).

`lint-legacy-names` now skips git-ignored files. Its exclusions named the build outputs we knew about (`dist/`, `.cache/`), which missed an artifact at a package root — dsh's `client.js` tripped it on `param-safe` back-compat reads inside the inlined runtime. Asking git is the general form of the question that path list was approximating. Verified it still catches a real violation in a tracked file, because a lint that silently looks at zero files is worse than one that fails.

## Gate status

`pre-pr.sh` green except two failures reproducible on `master` and unrelated to this branch: `sync.test.cjs`'s `--wsl outside of WSL` case (it unsets `WSL_DISTRO_NAME`/`WSL_INTEROP` but `isWsl()` also reads `/proc/sys/kernel/osrelease`, so it cannot pass on a real WSL machine — a real test defect worth its own PR) and a hermeticity diff caused by a live Claude Code session's commitments poller touching `~/.cues` mid-run. Doctor's stale-fork rows are the expected consequence of the version bumps.

## Known gaps

- **Inline notes float as an overlay** rather than splicing into the line the way terminal hosts do. Anchored and theme-aware, but a box above the text.
- **Firefox untested** — feature-detects `CSS.highlights` (140+) and degrades.
- **dsh is `0.1.0-rc`** and advertises breaking changes; the two contracts this builds on are annotated frozen upstream.


wkasekende:
## Coexistence with the chrome extension — was broken, now fixed

Tested rather than reasoned about, and the answer was **no, it did not work**. With both the extension and the plugin installed, the extension — keyless on a fresh profile — won the race and wrote its own error over the plugin's answer:

```
the capital of iceland is [OpenCues: no API key — open the extension popup]
```

An error about a credential that host does not even need, since it routes through the app's own model.

**Why no existing guard caught it.** A chrome content script runs in an **isolated world**: its own `window`, so the page-level singleton each host uses to survive its own remounts is invisible across the boundary. What they share is the document — same textarea, same key events, same `CSS.highlights`. The prior note in `integrations/dsh/CLAUDE.md` said to "scope it off that origin", which would only ever have helped a developer running dsh on a known port, not a user with both installed.

**The fix** is `@opencues/core/page-ownership`. An embedded host calls `claimPage(hostName)` (sets `data-opencues-host` on `<html>`); a host attaching from outside calls `pageClaimedByOther(hostName)` and stands down. Two decisions worth stating:

- **A DOM attribute, not a global** — precisely because a global cannot cross worlds. That constraint *is* the design.
- **Read live at every action point, never cached at boot.** The extension injects at `document_end` while an embedded plugin boots later (framework mount + async config fetch), so the claim usually appears *after* the extension has initialised. A startup-only check would miss the common case; there is a boot-time fast path too, but it is the optimisation, not the guard.

Precedence is deliberately one-directional: the **embedded host wins**, because it knows its editor's real shape, is version-matched to it, and usually has an LLM route already, whereas a generic extension guessing at someone else's composer is the weaker claim.

### Verified both directions

- **Both installed:** `pageClaim: "dsh"`, the plugin's fill lands as `the capital of iceland is Reykjavik`, exactly one substitution, extension silent.
- **Extension alone, no claim:** **all 22 chrome E2E tests pass**, including the trust-gate, sensitive-field and site-filter security controls. This was the assertion that mattered — a guard that stood down too eagerly would have disabled the extension on every ordinary page.
- 10 new core unit tests, including that a host never defers to *itself* (otherwise a remount re-claiming would make it go inert) and that the disposer does not retract a claim another host has since made (which would hand the page to nobody and re-enable the double-driving).

### One diagnosis worth recording

The first attempt at this fix appeared not to work: `pageClaim` read `null` even though the marker was in both bundles. Cause was my own documented trap — port 3081 had the plugin installed as a **tarball**, frozen at pack time, and re-packing to the same filename does not reinstall (pnpm reports "Already up to date"). Needed a `remove` plus a distinct filename. Exactly the "a stale bundle is completely silent" failure this PR's `ensureFreshBundle` step exists to prevent.

### Also in this push

- A stale `inline-note.e2e.test.ts` expectation asserted the overlay contained the **cue name**; `inlineNoteText` has no notion of one, so it was never satisfiable by any runtime this repo has shipped. Being run-on-demand rather than a CI gate, it sat failing and made every future real regression in that file look like the same known failure. Now asserts what the paint guarantees: connector, cycle countdown, previews a destination rather than echoing input, and the cue actually resolved.
- Chrome `manifest.json` + `package.json` bumped **in lockstep** to 0.2.166 per the repo rule, since chrome `src/` changed.
- `scripts/check-npm-fresh-install.sh` run against a local `opencues-0.7.0.tgz` in a pristine `node:22` container (Docker): **ALL CHECKS PASSED**. The `cli=v0.7.0 repo=v0.6.1` warning is the documented tag fallback — no `v0.7.0` tag exists yet — and resolves when the release is tagged.

Remaining `pre-pr.sh` failures are unchanged and unrelated: the WSL-detection family (`sync --wsl`, `which`'s WSL row — both clear the WSL env vars but `isWsl()` also reads `/proc/sys/kernel/osrelease`, so they cannot pass on a real WSL box) and a load-dependent seed-configs flake that passes standalone.

Issue #54, opened 2026-06-02
LLM features dead on Cerebras+Groq: runtime sends Groq's `openai/gpt-oss-120b`   to Cerebras (model_not_found)


deng-diana:
## Summary
  On a setup with both `CEREBRAS_API_KEY` and `GROQ_API_KEY` set and no explicit
  `llm-provider`, **every LLM feature silently fails** — no word-cue highlights,
  no `_` blank fills. Not a rate limit, not a keymap issue. The runtime sends
  Groq/OpenRouter's model name `openai/gpt-oss-120b` to **Cerebras**, which only
  accepts `gpt-oss-120b` (no `openai/` prefix).

  ## Impact
  Likely hits **every Cerebras+Groq user with default config, out of the box** —
  all LLM-backed surfaces (word-cues, fluid-blank, transform-blank, answer,
  prompt-improver) produce nothing, with no user-visible error.

  ## Reproduce
  Runtime log (`debug-mode: on`):
  ```
  FluidBlank/TransformBlank: failed — provider error: Model openai/gpt-oss-120b
  does not exist (code=model_not_found) [llm=cerebras/openai/gpt-oss-120b]
  ```
  Direct API repro with the same Cerebras key:
  - `cerebras` + `gpt-oss-120b`        → ✅ 200
  - `cerebras` + `openai/gpt-oss-120b` → ❌ "Model openai/gpt-oss-120b does not
  exist"

  ## Expected vs Actual
  - **Expected**: auto-route picks cerebras → uses cerebras's own default model
  `gpt-oss-120b`.
  - **Actual**: provider = cerebras (correct), but model = `openai/gpt-oss-120b`
  (Groq's format) rides along → rejected.
  
  ## Root cause (best guess — maintainer to confirm)
  The codebase already knows this footgun:
  - `sources/build-sources.ts` `resolveFor` unpairs the global model from the
    provider — but the unpair only seems to fire when a **bucket/explicit**
    provider is set, not on the **auto-route** path.
  - `sources/fluid-blank-source.ts:586` literally comments
  "`openai/gpt-oss-120b`
    sent to Cerebras".
  - `blanks/answer.ts:62` + `blanks/prompt-improver.ts:68` hardcode
    `?? 'openai/gpt-oss-120b'` — same footgun.
  - Cerebras's correct default is `gpt-oss-120b` (`llm-provider.ts` CEREBRAS);
    Groq's is `openai/gpt-oss-120b` (GROQ).

  ## Environment
[...]

PR #396, opened 2026-08-17
fix(dsh): a cloned marketplace install shipped no browser half

wkasekende:
Docs only. Records the distribution surfaces found while submitting `@opencues/dsh` to the DeepSeek Harness ecosystem, so a future release does not have to rediscover them.

The useful finding is not the list of directories — it is that they split into two kinds, and only one kind ever needs action:

**Auto-collected from the `dsh-plugin` GitHub topic** — the in-GUI marketplace (CI every 2h), three awesome-lists (6h / daily timers), and three web directories. The topic is the entire mechanism, and it is the one dsh's own README endorses. Nothing to submit, now or per release.

**Hand-curated, submitted by PR:**

| List | PR |
|---|---|
| awesome-dsh-plugin (7.5k★) — the authoritative catalog | [#1508](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/pull/1508) |
| 0xsline/awesome-deepseek-harness (687★) | [#374](https://github.com/0xsline/awesome-deepseek-harness/pull/374) |
| Anil-matcha/awesome-dsh-plugin (927★) | [#26](https://github.com/Anil-matcha/awesome-dsh-plugin/pull/26) |

The first is load-bearing beyond its own site: [dsh-market](https://github.com/dsh-market/dsh-market) (the plugin market *inside* dsh, one-click install) reads its `plugins.json` daily and explicitly says not to submit there directly. So that single YAML entry is what reaches users browsing from inside the app.

**Also records what was deliberately not done.** Dominic789654's list requires a `dsh` topic rather than `dsh-plugin`, and GitHub caps a repo at 20 topics — `opencues/opencues` was already at 20, so `dsh-plugin` cost `gpt-oss`. Spending a second slot on a 123★ list is a judgement call about your marketing surface, so it is flagged rather than done quietly.

## One risk worth reading

The marketplaces install by **cloning the repo** and running `npm install --omit=dev --ignore-scripts`. `client.js` is a build artifact — gitignored, produced by `prepublishOnly` — so it is in the npm tarball and **not** in a git clone, and `--ignore-scripts` guarantees it will not be built either. A clone-based install would therefore get the node half with no browser half: the config route answers and nothing paints or fills.

The documented path (`dsh plugin add @opencues/dsh`) resolves the tarball and is unaffected, and their spec does say the npm package name is the install target — but the pipeline description mentions cloning too. Left **unfixed on purpose**: both mitigations are unattractive (commit 1.7MB of derived output, or ship a `prepare` script that `--ignore-scripts` also defeats) and the risk is unverified, so the note says to verify before choosing rather than acting on a maybe.


wkasekende:
## The risk in the original description is now fixed and verified

The first version of this PR documented a clone-install risk as *unverified and unfixed*. Both are now false, so the scope has grown from docs-only to a real fix. Retitled accordingly.

### It was real, and confirmed from source rather than prose

I read `lib/index.js` of DSH-Plugins-Marketplace instead of trusting its README. It clones the repo and runs `npm install --omit=dev --ignore-scripts`, with **`allowScripts=false` as the safe default**, then copies the result into the profile. Its own fallback message states the expectation outright: *"using the build artifacts committed in the repo."* They have already been bitten by this exact bug class by another plugin (their issue #54, where a package's content lived only in the published tarball) — which is why their npm-tarball fallback exists at all.

So `prepublishOnly` never ran, `--ignore-scripts` guaranteed nothing would build it, and a gitignored `client.js` was simply absent. The node half loads, the config route answers, **nothing paints or fills** — broken with no error anywhere.

### The fix

`client.js` and `default-opencues.md` are now **committed**. That is the opposite of what this repo does everywhere else — `integrations/chrome/dist/` stays uncommitted, deliberately — and the justification is narrow rather than a change of principle: chrome is not distributed by clone-scraping marketplaces, and dsh is.

Two things pay for committing derived output:

- **`minify: true`** — 1.71MB → **1.01MB**, halving the per-change cost in git history and shrinking what every dsh page load pulls.
- **`scripts/check-dsh-bundle-fresh.sh`** — a committed artifact one version behind its source is precisely the silent-staleness failure this codebase has been bitten by repeatedly. esbuild is byte-reproducible for identical inputs (verified: two consecutive builds hash identically), so the gate **rebuilds and diffs** — exact, not heuristic. It builds core + runtime first, because the bundle inlines their dist and a stale dist would otherwise let a genuinely stale bundle pass. Wired into `pre-pr.sh` and CI.

### Proven against the path that was broken

Reproduced the marketplace install exactly — `git clone`, copy the plugin directory into a profile, **no npm tarball and no build step** — on a machine with no `.cues` config and every OpenCues provider key stripped:

```
config: none found on disk — using 29 shipped default(s)
Resolver: built with 7 sources
the capital of iceland is _  → "the capital of iceland is Reykjavik"
i has three cats fix typos _ → "I have three cats"
```

### One follow-on catch

Committing the bundle made it visible to the legacy-names lint, which flagged the `param-safe` back-compat reads inside the runtime code esbuild inlines — minification strips the per-line allow markers those lines carry in source. The lint now skips on the **"generated … do not edit" marker rather than the path**, so the next committed artifact is covered without editing the script again. Verified it still fails on a real violation in a tracked source file, because a lint that quietly stops looking is worse than one that fails.

Also recorded: their CLI-hint detection reads the **repo root** `package.json`, whose name is `opencues`, not `@opencues/dsh`. The npm path is found via the install command in the root README instead — another reason not to depend on the npm path being the one taken.

### Gates

`pre-pr.sh`: 21 gates (the new one included), dsh bundle freshness green. The 2 failures are the same pre-existing WSL-detection pair that cannot pass on a WSL developer machine (`isWsl()` reads `/proc/sys/kernel/osrelease` in addition to the env vars those tests clear) — unchanged by this branch, and they pass on CI's Linux runners.

Bumped to **0.1.2** so npm carries the minified bundle and matches the repo; 0.1.1 was never broken, just out of step with what is now committed. **Needs a publish after merge.**

PR #398, opened 2026-08-17
fix(ci,cli): stacked PRs ran no CI, and one WSL predicate replaces seven

wkasekende:
Two gate bugs, both of which made a check **less trustworthy rather than louder** — plus two more found by running the gates on this change.

## 1. Stacked PRs ran zero CI

`pull_request: branches: [master]` matches the PR's **base**, so a PR stacked on another feature branch matched nothing and produced **no checks at all**. GitHub renders that as an empty check list, which is visually indistinguishable from a green one.

**#395 reached `mergeable / clean` with no CI behind it** and was very nearly merged that way. Retargeting its base to `master` did not help either: that is a `pull_request: edited` event, and `edited` is not in the default type set (`opened` / `synchronize` / `reopened`), so nothing re-triggered. It took closing and reopening the PR to get a run.

The `branches:` filter is removed — every PR runs regardless of base. A run on a stacked PR costs one run; not running costs a merge with no evidence. `workflow_dispatch:` is added alongside, because when #395 had no checks there was also no event that would produce any.

## 2. Two tests could only pass on non-WSL machines

`sync.test.cjs` and `which.test.cjs` simulated "not under WSL" by deleting `WSL_DISTRO_NAME` and `WSL_INTEROP`. Reasonable, and not sufficient: detection also reads `/proc`, which keeps reporting the truth on a real WSL box. Both therefore **failed for every WSL developer and passed on CI's Linux runners** — `pre-pr.sh` was red on every change regardless of the change, which is how a gate stops being read.

**The cause underneath was duplication.** There were **seven** near-copies of the predicate — `sync.cjs`, `which.cjs`, `install.cjs`, `openrouter-oauth.cjs`, chrome's `bin/install.cjs`, and four inline anonymous functions inside `doctor.cjs` — and they had already drifted:

- most read `/proc/sys/kernel/osrelease`, doctor's four read `/proc/version`
- **`openrouter-oauth.cjs` checked the env vars only**, so it answered `false` on a WSL machine whose shell had not exported them. `wsl.exe -- node …` spawns exactly that shell. That one was a real behavioural bug, not untidiness.

Detection now lives once in `packages/opencues-cli/src/lib/is-wsl.cjs`, checks both `/proc` files, and carries a test seam (`setWslForTests` / `resetWslForTests`) modelled on the one `@opencues/core` already uses for its CLI-binary probe. **All seven call sites converted in this pass**, per fix-the-class-not-the-instance.

`which.test.cjs` also gains the **positive control** the original pair lacked — asserting the WSL row *appears* under WSL. The old assertion would have passed just as happily if that row had been deleted outright; "absent" only means something next to "present". 7 new tests cover the seam, including that a non-boolean override clears rather than coercing.

## 3 & 4. Found by running the gates on this change

Neither is in the work above; both are in the dsh bundle machinery added last week.

- **The documented build command could produce a different artifact.** `pnpm --filter @opencues/dsh build` runs the package script directly and skips turbo, so it bundles whatever `core`/`runtime` dist is on disk — against a stale one it silently produced **987kB where the committed bundle is 1,011kB**. The safe form is `pnpm build --filter @opencues/dsh`, whose `dependsOn: ["^build"]` builds the inlined packages first. Corrected in the README and the gate's fix hint. The gate itself was right: it builds them before diffing, which is why it kept converging on the correct bytes.
- **Turbo cached that build with no recorded outputs** (`outputs: ["dist/**"]` matched nothing, since this package writes to its root). A cache hit restores nothing, so it could leave `client.js` stale or absent while reporting success — the same silent-staleness class the gate exists to prevent, arriving through the cache instead. Added an `@opencues/dsh#build` override with the real outputs.

## Verification

`pre-pr.sh` went from **4 failures to 2**, and the WSL pair is gone entirely (0 occurrences in the log). The 2 remaining are the pre-existing pair, unrelated to this branch: a load-dependent `seed-configs` flake that passes standalone, and `doctor`'s install-state warnings.

CLI suite 690/691 with that same flake. `which` and `doctor` smoke-tested by hand after conversion, since parsing is not running.

One incidental catch worth noting: the new `HOSTS` drift test failed at first because this checkout's `opencues-core/dist` predated the `dsh` host — source had it, build did not. That is the test doing its job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,13 +17,29 @@ on:
       - "docs/**"
       - "spec/**"
       - "LICENSE"
+  # NO `branches:` filter here, deliberately. `branches: [master]` matches the
+  # PR's BASE, so a STACKED pr — one based on another feature branch, which is
+  # how a dependent change gets reviewed — matched nothing and ran zero checks.
+  # An empty check list is visually indistinguishable from a passing one, so
+  # #395 reached "mergeable, clean" with no CI at all and was nearly merged
+  # that way. Retargeting its base to master did not help either: that is a
+  # `pull_request: edited` event, and `edited` is not in the default type set
+  # (opened / synchronize / reopened), so nothing re-triggered. It took a
+  # close-and-reopen to get a run.
+  #
+  # Running CI on every PR regardless of base is the correct default: the cost
+  # of a run on a stacked PR is one run, and the cost of not running is a merge
+  # with no evidence.
   pull_request:
-    branches: [master]
     paths-ignore:
       - "**.md"
       - "docs/**"
       - "spec/**"
       - "LICENSE"
+
+  # Lets a run be kicked by hand — the escape hatch that was missing when #395
+  # had no checks and no event would produce any.
+  workflow_dispatch:
 
 # Cancel older runs of the same branch when a new push lands.
 concurrency:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed — a busy machine could fail the shipped-script smoke tests (`@opencues/runtime` 0.31.1 → 0.31.2)
+
+`blank-scripts.test.ts` spawns the **real** shipped `brightness-blank.sh` and `volume-blank.sh` and asserts each returns a bare integer. On WSL those reach a platform backend — brightness goes out to Windows through PowerShell/WMI — which is ~2s idle and considerably worse with the rest of the suite running in parallel, so vitest's 5s default timed out intermittently and presented as a product flake.
+
+That latency is not ours to bound, and the assertion is about the script's **output shape**, not its speed. Both now carry an explicit 30s timeout, so a busy machine no longer decides the verdict.
+
 ### Fixed — the hermeticity gate cried wolf on a running host's own writes
 
 `check-test-hermeticity.sh` asks "did a test write to the user's real config?" but was answering a broader question: "did anything under `~/.cues` change?". A live Claude Code or OpenCode session in another terminal answers yes on its own — its commitments poller rewrites `.session-commitments.kick` every few seconds, and a blank that fires updates `.user-blank-state/`. Neither is a test escaping its sandbox, and the gate reported `HERMETICITY VIOLATION` either way, which trains people to skim past the one piece of output that must never be ignored.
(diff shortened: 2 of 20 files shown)


wkasekende:
## Correction to this PR's description

The description claims the unsafe build command "silently produced **987kB where the committed bundle is 1,011kB**". **Those are the same file.** `build.mjs` prints KiB and `ls` prints bytes: 1,011,171 / 1024 = 987.5. I compared two units and called it drift. Verified properly since — identical `sha256` either way, so no drift was ever observed.

What survives is weaker and still true: `pnpm build --filter @opencues/dsh` goes through turbo and guarantees the inlined `core`/`runtime` are current, so it is the better habit. It is insurance, not a fix for a demonstrated bug, and the gate comment now says exactly that rather than citing a measurement that does not exist.

**The turbo `outputs` override is unaffected** and independently justified: the "no output files found" warning was real, and a task cached with no recorded outputs restores nothing on a hit, which for a committed artifact means a cache hit could leave `client.js` stale while reporting success.

## Also added since the description was written

Three more machine-dependent failures, all the same family — a test that could not survive the machine it ran on:

- **`seed-configs` raced with itself.** The Windows helper-`.exe` compile staged at a fixed path in the Windows home dir, so two concurrent runs deleted each other's files mid-copy (TOCTOU: `existsSync` true → sibling unlinks → copy throws `ENOENT`). And the block had a `finally` with **no `catch`**, so an optional Windows-only optimisation became a hard failure of `opencues install`. Staging paths are unique per invocation now and the failure is caught. That is a real product bug, not just a test one — verified by running the affected suites six-up and four-up concurrently, clean each time.
- **The hermeticity gate cried wolf.** It asks "did a test write the user's config?" but answered "did anything under `~/.cues` change?" — and a live Claude Code session in another terminal rewrites `.session-commitments.kick` every few seconds. It reported `HERMETICITY VIOLATION` for that, which teaches people to skim past the one output that must never be skimmed. Ephemeral host-owned state is now excluded; the config surface stays fully watched, so PR #41's class is still caught (114 noise lines dropped, 67 config lines still watched, hand-verified).
- **The shipped-script smoke tests** spawn the real `brightness-blank.sh`, which on WSL reaches Windows via PowerShell/WMI — ~2s idle, worse under parallel load, against vitest's 5s default. Explicit 30s timeout; the assertion is about output shape, not speed.

## Result

`pre-pr.sh` is **21/21** and `opencues doctor` reports **no issues found** — both firsts. Every host rebuilt to runtime 0.31.2 / core 0.46.0, the chrome `/mnt/c/` mirror synced, and the two `inherit` fall-through sentinels removed from `OPENCUES.md` with routing verified unchanged (`cerebras · gemma-4-31b` on all three buckets).

PR #414, opened 2026-08-25
README hero: the slot plays, and every word in it is the product's

wkasekende:
The hero placeholder now plays. **Draft — put up so the content can be judged
before more is built on it.**

## What it shows

Two four-ask chains, back to back, 53s. Every answer was **captured from the
running product**, not written: the runtime's fused transform-blank prompt
through the routed provider at its own settings (`cerebras gemma-4-31b,
temperature 0, seed 42`), each ask typed onto the answer before it.

**Claude Code**
`build a website improve prompt _` → `add a paragraph about security _`
→ `add a paragraph about html canvas _` → `translate to japanese _`

**Gmail**
`draft a resignation email body _` → `make up the details _`
→ `add emojis _` → `translate to japanese _`

The first grows to three paragraphs on ONE span and then carries all three
across. The second has the model place **eight** emoji, and the translation
keeps them. Both are the runtime's own chaining behaviour —
`findChainableLlmDef` extends the def rather than clobbering it — which is
exactly what a hand-written demo gets wrong.

Configs and captures live in the website repo's artifact kit
(opencues-web#40), which is where the film system lives. The captures sit
beside the configs, so the gate holds the film to the product's words and
re-capturing is the only way to change them.

## The embed

An animated WebP on a **relative path**, the same way `assets/supports.svg` is
already embedded. GitHub renders it inline as an ``, so there is no upload
step and no `github.com/.../assets/` URL to mint.

Encoded with the kit's own ruled settings — `-lossless 1 -compression_level 2
-loop 0`. A `q:v 62` pass came back visibly blocky on the flat dark ground and
smeared the `auto mode on` row; on UI text, size is the cheaper thing to spend,
which is the kit's own note. 2.1 MB, 385 frames, 960px wide.

## Deliberately unfinished

- **53s is about double** what a hero should run. Compressing the spinner turns
  and the settle holds gets each chain to ~15s without dropping any ask.
- **No third segment** for the note system yet.
- **No theme cutting inside a chain.** `swap` during a blank chain does work —
  `cc → slack → x` builds and renders — but three hosts need individual `type`
  tuning to share one panel height.

The other two placeholders (quickstart, feature tour) are untouched.


wkasekende:
Superseded by #418, which filled all three README video slots with lighter block-system films. Closing unmerged.

PR #422, opened 2026-08-27
README films: the runtime does what they claim, and the date cue cannot rot

wkasekende:
Two of the three README films assert things the product does not do. Both re-rendered.

## Slot 1 wiped when it should fill

The hero's closing beat replaced the whole line with the answer. `fluid-blank-source.ts` starts at **FILL** and only wipes when the host declares the field `disposable`, or declares `singleLine` **and** the buffer holds nothing but the query. No adapter in the product sets either — they are browser field shapes — so in a terminal the ask stays and only the `_` is replaced.

```
ffmpeg command to convert a video to web-ready mp4 ffmpeg -i input.mov -vcodec libx264 …
```

with only the command selected. I had taken `MODE: WIPE` from the capture tool's output and skipped the sentence under it saying the runtime ignores the model's advice.

## Slot 1, taken further — taught the runtime to do what the video claims

Downgrading the video to FILL fixed the false claim, but the point of this PR is that the video should be *true*, not merely non-false. `bufferIsExactlyTheLookup` (buffer is provably nothing but the query, no paragraph break) was already sitting in `fluid-blank-source.ts`, deterministic and data-loss-free — it was just gated behind `ambient.singleLine`, a browser field-shape flag no native host adapter ever sets. Dropped that requirement and paired the floor with the model's own `MODE` vote instead: the floor alone can't distinguish a terse lookup from a compact factual sentence with a mid-span gap (both are span=whole-buffer), only the model's shape classification can, and the floor is what makes trusting that vote safe — it's already proven nothing else in the buffer is at risk before the vote is even read. Not the model-decides-alone heuristic that flattened real documents and got retired in `f62dcd28`; the floor is checked first, unconditionally.

Verified live against cerebras + groq (temp=0, seed=42), not just mocked: both correctly WIPE the ffmpeg case and `capital of france _`.

That live validation run also caught a second bug, unrelated to and pre-dating this change: FILL's splice always replaced just the `_` character. FluidBlank's own ANSWER RULE 5 tells the model to answer a mid-sentence gap with the full restated clause, and splicing that at `_` alone duplicates the surrounding words — confirmed for real, `There are _ continents` → `There are 7 continents` → spliced result `There are There are 7 continents continents`. New `findRestatedClauseSpan` detects the shape deterministically (string check, no model-vote trust) and widens the splice to the verified span; still tagged FILL.

`@opencues/core` 0.54.0 → 0.55.1. Tests: 6 new unit tests + 2 new resolver-level scenario tests (the ffmpeg WIPE journey end-to-end including cycling back to the original ask, and the restated-clause splice with the real captured model output run through the real resolver splice primitive — not a mock). Full design: `docs/architecture/blank-sources.md` § WIPE gate.

## Slot 1, the actual re-render — off a different mechanism than first shipped

Re-rendered twice more after the above. First re-render used `films/medium-film.html`'s film27 (a **data-film catalog entry**), which `build.py`'s `film()` wraps in a title + description unconditionally — the catalog page needs them, there's no flag to omit them. That baked "film 27 · a chain of asks..." into every frame — not what "just the terminal on transparency" claimed. Second re-render also fixed a width mistake (`--width 780`, chasing the old file's exact pixel dimensions instead of this README's own documented recipe: `--width 520 --scale 2`, "all three films are cut at 520").

The actual fix: the README's *other* title-free assets (`block-hn-full.png`, `block-gmail-520.webp`, `block-x-520.webp`, `block-metric-state.png`) were never captured off the film catalog at all — they're bare `.term`/`.lt-xxx` blocks on the **data-live** engine, no film wrapper, nothing to strip. `opencues-web`'s new `artifact-kit/films/readme-hero-live.py` follows that precedent: one `.lt-cc` block, a `data-live` steps sequence carrying the identical chain (same words, same product run — cerebras gemma-4-31b, temperature 0, seed 42, never rewritten), captured directly. Verified with still captures at three points before committing to the cloud render: mid-typing (no title), mid-chain (note showing correctly), and the ffmpeg WIPE ending (note matches film27's own captured `"2 | ffmpeg command..."` exactly).

Final: 1040×1266 (was 1040×1182 on film27, close), 219KB (was 772KB with the catalog chrome baked into every frame). Dropped the walk-back cycling beat from this cut (out of scope) — the growing chain and the WIPE ending are both there.

Known tradeoff, accepted: `data-live` reserves the tallest state's height across the whole clip with no opt-out, so early frames show more empty terminal than film27's own `NO RESERVE` blank-chain program did — the content spans one line to a full paragraph-grown Japanese translation, so the range is real.

`films/film27` itself is untouched apart from the WIPE-mode fix — it still tells the same story, theme-cycling across every host, for the site's own films catalog page. This is the purpose-built, title-free cut of the same content for the README specifically.

## Slot 2's date cue would have gone false

`let's ship it Thursday the 19th` carries no year, so the check resolves the **upcoming** 19th against the clock:

| now | tip |
|---|---|
| 2026-01 | the 19th is a **Monday** |
| 2026-04 | the 19th is a **Sunday** |
| 2026-06 | the 19th is a **Friday** |
| 2026-10 | the 19th is a **Monday** |
| 2026-12 | the 19th is a **Saturday** |

The film baked in "Friday", so it was correct in one month of six and a false claim in a README for the rest. A rendered webp cannot self-correct.

It now reads `freeze is Wednesday, September 3 2027` → `⚠ 2 | the 3rd is a Friday, not Wednesday`, verified stable against a clock moved across six months. The explicit year pins the resolution — the runtime has a test for that, added after a July-2026 time-bomb false positive. September 3 2027 is a Friday.

**Rule for any future date demo: write the year.** Without one the demo is only true in the month it was made.

## Rendering

Both cut in the cloud via `lambda/invoke.cjs`. Worth noting for anyone following the kit docs: `lambda/block-shot.cjs` is the copy that runs *inside* the sandbox — calling it from a workstation runs the frame probe locally and only ships the sheets.

Hero 1040x1182, 389 frames, 44.2s, 807KB. Cues 1040x294, 102 frames, 18.2s, 103KB. Slot 3 is untouched.

Kit side, with the film-runner fix and the gates: opencues/opencues-web#41

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

https://claude.ai/code/session_01GYaAxCvS8C8tbmc7VxEMC8


wkasekende:
The hero's answers now GLIMMER in (`f18a115c`): since v0.7.8 every substitution lands with the 140ms blink + 70ms churn easing 0.45 → 0.15 at 900ms on every host, so the hard-cut landings this PR shipped were already one release behind the product. The kit derives the landing on every blank `resolved` (opencues-web `fix/readme-film-truth`, engine `glim` step + derive.py `_glim` — see opencues/opencues-web#41); this is the re-render off the same two joined solo cuts and the same captured answers. 1040×1002 as before; 1.34MB vs 400KB — churn frames are high-entropy and never fold.

PR #431, opened 2026-09-04
feat(cc): pin to 2.1.236 (Anthropic stable) — S7 restored; 2.1.243+ blocked upstream

wkasekende:
## What

CC integration pin bump: **2.1.206 → 2.1.236** (Anthropic's `stable` dist-tag), tweakcc pin **`1545ff8` → `371a5c46c`** ("Prompts for 2.1.236"). Built and validated overnight on an isolated fork per the UPGRADING.md runbook — **the canonical `claude-cues` fork is untouched**; it upgrades via `opencues update claude-code` after this merges.

## Why 2.1.236, not 2.1.260 (latest)

**CC 2.1.243+ is currently un-patchable.** Those builds use Bun code splitting: a ~19KB entry stub importing ~1400 `chunk-<hash>.js` modules, each carrying precompiled bytecode. I verified our seam regexes survive the split (S1/S2/S3 all present in a single chunk on 2.1.259 — probed against every one of the 1819 archive modules), but the patch pipeline doesn't:

- upstream tweakcc **#978** (open): the 4.3.3 parse gate rejects the new ESM entry chunk, so every apply rolls back
- upstream tweakcc **#979** (open): repack on a code-split binary bloats 392MB → 694MB with duplicated chunks and applies no config
- unverified: whether Bun executes patched module TEXT over the module's stale bytecode — a silent-inert-patch risk

`2.1.242` is the last pre-split version. The ceiling + re-entry checklist is documented in `compat.json://code-split-ceiling`, UPGRADING.md § "Code-split ceiling", and REPAIR.md § 16 — when both upstream issues close, the runbook picks up from there.

## What the bump fixes/changes

- **S7 (RenderKick) is restored** — it was missing on 2.1.206, so pushes fell back to the ZWS-toggle; on 2.1.236 the seam hits and the clean explicit-kickRender path is back. S6 stays gone (polling fallback, as since 2.1.150).
- Two tweakcc 4.3.3-era regressions handled in `setup.sh` (full post-mortem: **REPAIR.md § 16**):
  1. `--apply` now requires `--yes` (new confirmation gate aborted every headless install) — § 8 passes it.
  2. tweakcc's new `assertPatchedBundleParses` gate `node --check`s the patched bundle — but CC's embedded JS uses Bun-only `using` declarations Node can't parse, so it **false-positives and rolls back valid patches** ("SyntaxError: Unexpected identifier 'n'", hit live). New **§ 4f** drops the call, anchor-verified like § 4e. Nothing is lost: our § 9 smoke *executes* the patched artifact under Bun, which is strictly stronger than a Node parse. `check-tweakcc-pin.sh` § 3b pins the disable.

## Validation (isolated fork, Linux x64)

- Full `setup.sh` install of 2.1.236: all steps green, S1/S2/S3 applied, `--version` runtime smoke passes on the patched binary.
- **14/14 CC-scoped agentic scenarios pass** against the patched binary (cycling core: 01/02/03/07/10/22/23/25/26/28/30/35/36/37) via the pty launcher on an isolated `OPENCUES_HOME`.
- `check-tweakcc-pin.sh` green with the new § 3b assertion; § 4e anchor verified present at the new tweakcc commit.

## Files

`compat.json` (both pins + tested array + code-split-ceiling note) · `setup.sh` (§ 8 `--yes`, new § 4f, comment pins) · `check-tweakcc-pin.sh` (§ 3b) · `UPGRADING.md` (ceiling section + seam-inventory header + pin refs) · `REPAIR.md` (§ 16) · `integrations/claude-code/README.md` + `package.json` (0.2.12, compat range 2.1.110–2.1.242) · root `CLAUDE.md` (install table + version map) · `CHANGELOG.md`

## After merge (your side)

1. `opencues update claude-code` (rewrites the fork pin + reinstalls + repatches), then restart `claude-cues`.
2. FYI: the dsh bundle-freshness gate passes on master post-qwen ("committed artifacts match their sources"), but regenerating locally produces symmetric line churn (nondeterministic build ordering) — I kept that byproduct out of this branch. No action needed unless you want the dsh inline bundle re-cut against core 0.57.0 with a version bump.

https://claude.ai/code/session_01XBtLCdx8JSPuhyzjRByAJj


wkasekende:
Belt-and-braces pass: 13/13 LLM-surface scenarios green against the patched 2.1.236 binary (transform/fluid pipelines, fluid-config flip, config-intent span preservation, three-trigger chain with no double-fire, [CURSOR] positional split, contradiction bill-split, session-contradiction, _-history walking, trailing-newline span trim, selector-satellite _-cycle, replace-parse bounded splice). Combined with the earlier cycling-core run: **27/27 scenarios on 2.1.236**. Skipped as documented-unrunnable on this host regardless of version: sentence-cue trio (shipped cue is `not-on-host: claude-code`), 119 (stocks blank shadows the input), 121 (opt-in, covered on opencode in the qwen validation).

PR #441, opened 2026-09-06
docs: semantic tips proposal — the tips pack as a watchlist on the rules matcher

wkasekende:
Plan only, no code.

The tips system is static token matching, which is the right zero-latency floor and the wrong ceiling: it cannot match situations ("start over", "it forgot the plan"), cannot carry expertise catalogues, and cannot steer between tips. The rules watchlist already does what we want for a different catalogue. This proposal generalises the session-contradiction matcher into a `WatchlistSource`, feeds it the host's tips packs as a second catalogue (a `when:` line per entry, a token budget instead of the 24-entry cap), keeps static matching as the floor, and makes a deterministic bench the ship gate.

`docs/architecture/semantic-tips.md`: what exists exactly, the design (engine, pack field, budget, rail position, bench), phases 0-6 with proof per phase, risks, open questions.

Grounded in the 2026-09-06 research (website repo, `internal/research/claude-code-cues/`).

https://claude.ai/code/session_01Aa52PYn55dTUfUWm3n3rEv

### DIFF
--- a/docs/architecture/semantic-tips.md
+++ b/docs/architecture/semantic-tips.md
@@ -1,0 +1,193 @@
+---
+last_updated: 2026-09-06
+status: PROPOSAL — not built. Ruled in principle by Wilfred 2026-09-06 ("what we want is more like the rule system ... takes the available tips as context which is then used to dynamically match").
+---
+
+# Semantic tips — the tips pack as a watchlist
+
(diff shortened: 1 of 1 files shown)


wkasekende:
Superseded by #442.

PR #479, opened 2026-09-18
feat: data tables answer a `_` with no LLM at all — the tables built-in, the data policy, the table verdict stacked on the route; `table-lookups-mode` (off by default)

wkasekende:
Stacked on #478 (device leg). Reimagined-plan item 3: data tables with no LLM at all.

## What ships
- **`tables` built-in** (`packages/opencues-runtime/src/blanks/tables.ts` + `defaults/blanks/tables/BLANK.md`): eight offline tables in one shape-gated blank. `unicode for em dash _`, `hex for tomato _`, `rgb for #1e90ff _`, `http status for not found _`, `mime type for png _`, `default port for postgres _`, `convert 5 miles to km _`, `calc 17 * 23 _`, `atomic number of gold _` (+ `atomic mass of`, `boiling point of`, `melting point of`, `ph of`). Data is bundled (`tables-data.ts`, generated by `scripts/gen-tables-data.mjs`); the calculator is a shunting-yard evaluator, no `eval`. `convert` / `calc` claim the `_` only when a number follows, so `convert this to markdown _` stays a rewrite. Works with `decisions-provider: off`.
- **`decisions/data-policy.ts`** (security-audit row #32 tier 2, the same rule as `device-policy.ts`): the 19 table ids a leg may name (the eight tables + six country facts on the existing `countries` blank) and the grammar that captures the ARGUMENT from the segment holding the `_` (`what port does postgres use _` → `postgres`, `how acidic is lemon juice _` → `lemon juice`), floored (length, control chars, URL structure; a numeric alphabet with no call / member syntax for the expression tables). The model only ever picks an id.
- **Wiring**: `DecisionLegs.table` (optional) + `UnderscoreRouting.table`, stacked on the `_` route request; resolver → `resolveDataInvocation` → `BlankFill.fillFromDecision`, which now PROBES the table (`lookupTable` / `countryFactAvailable`) before it fills — a hit runs the shape path with no chat call, a miss declines and the pass falls through to the fan-out as it would have. A table stands over a `lookup` route (the chat path it replaces), never over settings / transform. Bridge (chrome) carries the leg. Emits `resolver.table`.

## Tests
- `tables.test.ts` (every lookup + the evaluator's precedence / unary minus / no-eval), `data-policy.test.ts` (18 phrasings → keyword + argument, segment scoping, the floor), tables routing journeys in `blank-fill.test.ts` against the REAL BLANK.md (12 positives, 4 rewrite-shaped negatives, prior-prose survival), resolver table block in `resolver.route.test.ts` (lookup route yields, settings route wins, policy rejects, miss falls through), probe scenarios in `device-decision.scenarios.test.ts`. Registry drift lists + chrome collision-guard test updated.
- `bash scripts/pre-pr.sh`: 20/21 green; `doctor` only reports the pre-install stale-bundle rows.

## Measured (private package bench, end to end through the policy and the runtime's tables)
37/37 plain-phrased lookups answered, 0/12 traps fired, alone and stacked on the route; route decisions unchanged on 399 `_` (routed 303 → 304, right 296 both ways; +1.2k tokens, +20 ms on the `_` request).

## Upgrade path
`opencues install claude-code` (runtime + a new `defaults/blanks/tables/` seeded by seed-configs) · chrome: rebuild + `opencues sync chrome` + reload · the private package ≥ the commit with the `table` leg for the plain-phrasing path (keyword forms need nothing).

https://claude.ai/code/session_01FRepZa2uPEg4WHabfjvbUc


wkasekende:
Second commit: `table-lookups-mode: off | on`, **off by default**. Off leaves the `tables` blank unregistered AND drops the table question from the `_` route request (`RouteContext.tables`, over the bridge too), so it costs no tokens; on registers the blank and stacks the question (+1.2k tokens, +20 ms per `_`). Settings menu (Blanks), doctor, defaults/OPENCUES.md.

